PR #1, opened 2020-05-06
Add initial version of multiple choice interactive and its authoring

pjanik:
[#172594435]
[#172103931]

This is the initial implementation of multiple-choice questions. It's functional, both runtime and authoring. Tests are in progress, I'll open a separate PR with them. But I thought it might be worth getting some feedback earlier.

Authoring demo: https://authoring.staging.concord.org/activities/20635/pages/307397/edit
Runtime demo: https://authoring.staging.concord.org/activities/20635/pages/307397

I've added `useLARAInteractiveAPI` hook that can be reused by more question types. I didn't try to add there all the features of Interactive API. I think it's pretty small and readable now, so other projects can just take it and extend it if necessary. I guess that if we try to handle every possible option that API offers, the code will get much bigger, we'll be returning tons of options, and so on. Not sure if it's worth it. Also, I think that it'll be difficult to understand some concepts without documentation anyway (e.g. linked interactives or global state). 

aspectRatio is a bit problematic. We can't guess aspect as it'll depend on number of choices or prompt length. And forcing authors to set it manually for every question seems annoying. So, the interactive tries to guess it itself and report back to LARA. Fortunately, LARA accepts subsequent calls to "supportedFeatures" fine. You can test it by adding more choices in authoring. We could more explicit API for it though.

I could use react-jsonschema-form even in the runtime - generate schema using an authored state. But it seems like overkill and it can make styling and specific/custom features more problematic long-term.


scytacki:
A quick initial comment based on your description above. Instead of aspect ratio I think it would work better if you used the 'height' option in the interactive api. I don't think we get this documented in the interactive api documentation site, but there is a way the interactive can ask to be a specific height instead of asking for an aspect ratio.

scytacki:
This could be a separate PR, but it looks like it doesn't handle reporting on the state at least in the LARA report: https://authoring.staging.concord.org/activities/20635/summary/c03256eb-cb47-4b44-b92d-284fce7233dc

pjanik:
@scytacki, thanks, I didn't know about height option. I'll check it.
I'm assuming that reporting is a separate story. I asked in PT story about the preferred approach, we can discuss it there.

src/multiple-choice/components/app.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+  useEffect(() => {
+    // Measure the main app container on every re-render and update aspect ratio. Note that useLARAInteractiveAPI
+    // takes care of its updates and will notify LARA about it using iframe phone.
+    const width = container.current?.offsetWidth;
+    const height = container.current?.offsetHeight
+    if (width && width > 0 && height && height > 0) {
+      setAspectRatio(width / height);
+    }
+  });
```
scytacki, line 24: I guess since this useEffect doesn't have a list of dependencies it will run on every render. Will the `useRef` trigger a re-render if the interactive is embedded in a responsive layout page and the page is resized?

pjanik: Yes, this will run on every render. So, if some component is updated, user types, clicks, and so on. Usually, that's actions that cause size change (e.g. when the user adds a new choice in the form). It's a little heuristic, but I think that it should be working in most cases (a case where it won't work - user trying to resize textarea, as that doesn't trigger state change and rendering).

Re iframe resizing in the responsive layout - no, useRef doesn't monitor anything and doesn't trigger any renders. 

I thought this question would be used in the assessment block usually. But I think that interactive should always report its own aspect ratio and it shouldn't really care about size of the page or iframe. In this case - the question content aspect ratio. So, I don't think that page/iframe size should matter here. Even if the iframe is too small to display interactive completely and there're scroll bars, the aspect ratio will be reported correctly - I tested that and that's how it works while we're authoring and adding new choices. I guess for some milliseconds interactive is larger than LARA iframe, but interactive immediately sends new aspect ratio, LARA accepts that, and in the end we can see that height is increased and there're no scrollbars.

We could simply add document/window.resize handler here. But I think that page/window resize won't change the aspect ratio that we calculate (or at least shouldn't).

scytacki: Yeah it is an edge case that this would be added to the interactive box and thus affected by responsive layout. However we should handle it because it is possible.
If you switch to using height api instead of aspect ratio this will also simplify things.
But I suspect in either case the height or aspect ratio will want to change when the window is resized. For example a large prompt that spans multiple lines will reflow when the page is made more narrow or wide. So that should trigger a height or aspect ratio change.

scytacki: I made a test here to see what happens: https://authoring.staging.concord.org/activities/20598/pages/307321
It is interesting how the interactive starts to get shorter instead of longer as the page width is shrunk.  I guess that makes sense because the aspect ratio doesn't change so with less width there is less height too.

scytacki:
The new testing approach looks good to me.

scytacki:
This PR is getting big though so what do you think about merging it and starting another one for additional changes?

pjanik:
Sure. I merged reviewed testing into this one, but yea, it makes further reviews more difficult.

PR #2, opened 2020-05-07
Setup unit and integration/Cypress testing

pjanik:
1. Iframe-phone mock.
1. Unit testing (mostly hook).
1. wrapper.html - small page to wrap selected question-type in an iframe ane expose iframe-phone to window namespace.
1. Cypress tests using wrapper.html testing real-world iframe communication. I was thinking what approach would be the best for the integration test and this one is self-contained (we don't depend on real LARA instance) and I think it's testing integration quite well.


cypress/integration/multiple-choice.test.js
```diff
@@ -0,0 +1,124 @@
...
+      cy.getIframeBody().find("#root_choices_0_correct").click();
+      cy.getIframeBody().find(".btn-add").click();
+      cy.getIframeBody().find("#root_choices_1_content").clear();
+      cy.getIframeBody().find("#root_choices_1_content").type("Choice B", { force: true }).then(() => {
+        expect(receivedState.version).eql(1);
+        expect(receivedState.prompt).eql("Test prompt");
+        expect(receivedState.multipleAnswers).eql(true);
+        expect(receivedState.choices[0].content).eql("Choice A");
+        expect(receivedState.choices[0].correct).eql(true);
+        expect(receivedState.choices[1].content).eql("Choice B");
```
scytacki, line 119: My cypress is rusty. If receivedState hasn't been updated yet, and this expectation fails, will cypress automatically retry the whole block?  

I know it does that in some cases

pjanik: I think it re-runs cypress matchers like `.find` or `.get`. Not sure about these lines, but I doubt so, as it's just mocha/chai. So yes, I'm cheating here. I'm explicitly relying on the fact this promise is resolved later than iframe-phone message handler. I was testing that multiple times and it was consistent. But now I think I have an idea of how to do it better and more reliable. I'll test and push in the next commit.

pjanik:
@scytacki, I pushed a new commit. It extracts iframe-phone helpers and makes them more bulletproof. Now, Cypress will wait for the iframe-phone message if it's not available yet. I think it should be easier to use too.

PR #3, opened 2020-05-08
"height", extraInstructions, authoring and runtime unit tests

pjanik:
1. It replaces `aspectRatio` with "height" message and adds "resize" event listener, so responsive layout should be handled now.
2. `extraInstructions` added.
3. Adds missing unit tests:
Screenshot 2020-05-08 at 19 04 16

### DIFF
--- a/src/multiple-choice/components/app.tsx
+++ b/src/multiple-choice/components/app.tsx
@@ -4,24 +4,31 @@ import { Authoring } from "./authoring";
 import { Runtime } from "./runtime";
 
 export const App = () => {
-  const [ aspectRatio, setAspectRatio ] = useState(1);
   const interactiveAPIConfig = {
     interactiveState: true,
     authoredState: true,
-    aspectRatio
   };
-  const { mode, authoredState, interactiveState, setInteractiveState, setAuthoredState } = useLARAInteractiveAPI(interactiveAPIConfig);
+  const { mode, authoredState, interactiveState, setInteractiveState, setAuthoredState, setHeight } = useLARAInteractiveAPI(interactiveAPIConfig);
   const container = useRef<HTMLDivElement>(null);
 
-  useEffect(() => {
-    // Measure the main app container on every re-render and update aspect ratio. Note that useLARAInteractiveAPI
-    // takes care of its updates and will notify LARA about it using iframe phone.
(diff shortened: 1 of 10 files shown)


src/shared/hooks/use-lara-interactive-api.ts
```diff
@@ -34,6 +34,10 @@ export const useLARAInteractiveAPI = (config: IConfig) => {
     phone.current?.post("interactiveState", state);
   }
 
+  const handleSetHeight = (height: number) => {
+    phone.current?.post("height", height);
+  }
```
scytacki, line 39: Is it possible for setHeight to be called before the phone is ready?
If so, should this cache the height and then set it as soon as the phone is ready?

src/multiple-choice/components/app.tsx
```diff
@@ -4,24 +4,31 @@ import { Authoring } from "./authoring";
...
+  useEffect(calcHeight);
+
+  useEffect(() => {
+    // It's necessary to wrap calcHeight in a closure function. Note that calcHeight will be recreated on every render.
+    // This will break the cleanup function which would try to remove wrong reference. resizeHandler is created
+    // only once, as useEffect has empty dependencies list.
+    const resizeHandler = () => calcHeight();
+    window.addEventListener("resize", resizeHandler);
+    return () => window.removeEventListener("resize", resizeHandler);
+  }, []);
```
scytacki, line 31: I'd guess this will be a shared pattern between most of these components. So eventually this should be moved into the shared folder somehow.

src/shared/hooks/use-lara-interactive-api.ts
pjanik: I'm pretty sure iframe-phone takes care of that. It queues messages that are posted before the handshake and sends them after the connection is established.

src/multiple-choice/components/app.tsx
pjanik: Right, most of the App component will be reused. I'll think how to do it. I guess it can be a hook actually.

src/shared/hooks/use-lara-interactive-api.ts
scytacki: Yup that makes sense.

PR #5, opened 2020-05-11
Multiple choice: add report view + tests

pjanik:
[#172726949]

I did not use "No response" text, as then teachers wouldn't know if the question is multiple choice or open response. We can change it when reporting gets more advanced.


src/multiple-choice/components/app.tsx
```diff
@@ -30,10 +30,11 @@ export const App = () => {
     return () => window.removeEventListener("resize", resizeHandler);
   }, []);
 
+  const report = mode === "report";
```
scytacki, line 33: No need to change this now. But perhaps a good pattern for these components is to define 3 boolean variables:
```
runtimeMode, authoringMode, reportMode
```
And then logic like below will be consistent and concise.

PR #6, opened 2020-05-12
Add open response interactive (runtime, authoring, reporting + tests)

pjanik:
It follows the same pattern. There's a new hook `useAutoHeight` and some styles are shared now. I was considering creating something like `BasicAuthoring` component that would only accept schemas, but I'm not sure if it's worth it at this point and how long we can use just raw react-jsonschema-form. I might try to do it while working on fill in the blank question.


src/open-response/components/authoring.tsx
```diff
@@ -0,0 +1,84 @@
+import React from "react";
+import Form, { IChangeEvent } from "react-jsonschema-form";
+import { JSONSchema6 } from "json-schema";
+
+import "../../shared/styles/boostrap-3.3.7.css"; // necessary to style react-jsonschema-form
+import css from "../../shared/styles/authoring.scss";
+
+// Note that TS interfaces should match JSON schema. Currently there's no way to generate one from the other.
+// TS interfaces are not available in runtime in contrast to JSON schema.
```
scytacki, line 9: @pjanik Have you looked at this: https://github.com/YousefED/typescript-json-schema ?

pjanik: No, I haven't seen it before. It's interesting. It would require some additional build step. I think this library would be perfect to add additional runtime validation of data in some of our projects.

In our particular case, we use JSON Schema to generate forms and need more information than TS interfaces provide (e.g. titles). I can see that this library supports some annotations in comments (e.g. to add descriptions and titles), so it could still work. Since we have only a few simple schemas now, I'm not sure if it's worth it yet. It could be a research spike to check this library (and/or alternatives) and try to integrate it nicely with build system.

The opposite direction could be interesting in this case (json schema -> ts interface), connected with dynamic generation. Someone has tried:
https://github.com/jdmichaud/jsonschema-loader
But it's only webpack loader (and seems like a dead experiment). Not sure how well this would work with IDEs and e.g. jest tests that don't use webpack. Generally, this seems less likely to work.

scytacki: Yeah isn't worth it yet especially since we'd have to figure out the build step.

It is likely that we are going to use this approach of annotated TS to implement the new portal api. In that case we'd use the annotations to generate the ORM for the database. And the annotations to generate the graphql definition for the api. So perhaps once we have more experience with that, then it'll be easier to see if this approach is good or not.

PR #7, opened 2020-05-15
172104057 fill in the blank

pjanik:
This PR adds fill in the blank. Authoring isn't based only on keywords parsing, we're adding dynamically new input fields when needed. 

Once the question is authored, the runtime doesn't depend on the current keyword regexp. So, we can later change [blank-<ID>] to whatever string we want and it shouldn't break old questions.

I might extract authoring component to something like BaseAuthoring later, as it seems all the questions so far are fine with default React-JSONSchema-Form + optional state processing (to generate choice IDs or blank fields options).

Authoring demo:
https://authoring.staging.concord.org/activities/20643/pages/307426/edit
Runtime demo:
https://authoring.staging.concord.org/activities/20643/pages/307426
LARA student reporting (summary) seems to be broken when there's managed interactive that saves state, so can't show report for now.

### DIFF
--- a/src/multiple-choice/components/runtime.tsx
+++ b/src/multiple-choice/components/runtime.tsx
@@ -69,7 +69,7 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
           authoredState.choices && authoredState.choices.map(choice => {
             const checked = selectedChoiceIds.indexOf(choice.id) !== -1;
             return (
-              <div key={choice.id} className={report && getChoiceClass(choice, questionScored, checked)}>
+              <div key={choice.id} className={report ? getChoiceClass(choice, questionScored, checked) : undefined}>
                 <input
                   type={type}
                   value={choice.id}
--- a/src/multiple-choice/components/runtime.tsx
+++ b/src/multiple-choice/components/runtime.tsx
@@ -13,19 +13,6 @@ interface IProps {
   report?: boolean;
 }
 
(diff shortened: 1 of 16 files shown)


src/multiple-choice/components/runtime.tsx
```diff
@@ -69,7 +69,7 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
           authoredState.choices && authoredState.choices.map(choice => {
             const checked = selectedChoiceIds.indexOf(choice.id) !== -1;
             return (
-              <div key={choice.id} className={report && getChoiceClass(choice, questionScored, checked)}>
+              <div key={choice.id} className={report ? getChoiceClass(choice, questionScored, checked) : undefined}>
```
scytacki, line 72: Not sure it is worth it, but you could move `getChoiceClass` into the Runtime function so getChoiceClass would have direct access to the questionScored and report variables.  Then this logic could be moved out of jsx.

src/shared/hooks/use-delayed-validation.test.ts
```diff
@@ -0,0 +1,21 @@
+import React, { useRef } from "react";
+import { renderHook, act } from "@testing-library/react-hooks";
+import { useDelayedValidation } from "./use-delayed-validation";
+import Form from "react-jsonschema-form";
+
+describe("useAutoHeight", () => {
```
scytacki, line 6: This looks like a copy and paste error.

pjanik:
Thanks. I've fixed both issues in the last commit, so I'm merging it in.

PR #8, opened 2020-05-19
Scaffolded question

pjanik:
This is an initial implementation, no tests yet. But I think it makes sense to review it to see how it's done and get some feedback, as it's demoable now. Authoring uses react-jsonschema-form custom fields to provide iframe-based authoring, and this extension was pretty smooth. Available interactives are hardcoded for now.

Demo:
- authoring: https://authoring.staging.concord.org/activities/20649/pages/307447/edit
- runtime: https://authoring.staging.concord.org/activities/20649/pages/307447
- report: https://authoring.staging.concord.org/activities/20649/summary

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,6 +107,9 @@
     "webpack-dev-server": "^3.10.3"
   },
   "dependencies": {
+    "@fortawesome/fontawesome-svg-core": "^1.2.28",
+    "@fortawesome/free-solid-svg-icons": "^5.13.0",
+    "@fortawesome/react-fontawesome": "^0.1.9",
     "iframe-phone": "^1.2.1",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
--- a/src/scaffolded-question/components/iframe-authoring.tsx
+++ b/src/scaffolded-question/components/iframe-authoring.tsx
@@ -58,10 +58,6 @@ export const IframeAuthoring: React.FC<FieldProps> = props => {
       mode: "authoring",
       authoredState
     });
-    phone.post("initInteractive", {
-      mode: "authoring",
-      authoredState
-    });
   }
 
   useEffect(() => {
--- a/src/scaffolded-question/components/iframe-authoring.tsx
+++ b/src/scaffolded-question/components/iframe-authoring.tsx
@@ -1,0 +1,94 @@
+import React, { ChangeEvent, useEffect, useRef, useState } from "react";
+import { FieldProps } from "react-jsonschema-form";
+import { IframePhone } from "../../shared/types";
+import iframePhone from "iframe-phone";
+
+import css from "./iframe-authoring.scss";
+
+// Later, this will be replaced by LARA Interactive API call that returns all the available managed interactives.
+const availableInteractives = [
+  {
+    url: "",
+    name: "Select an interactive"
+  },
+  {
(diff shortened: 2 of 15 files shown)


src/scaffolded-question/components/iframe-authoring.tsx
```diff
@@ -0,0 +1,95 @@
...
+      setIframeHeight(newHeight);
+    });
+    phone.post("initInteractive", {
+      mode: "authoring",
+      authoredState
+    });
+    phone.post("initInteractive", {
+      mode: "authoring",
+      authoredState
+    });
```
kswenson, line 64: Is calling "initInteractive" twice necessary?

scytacki:
This looks really good. I'm amazed how simple the sub interactive stuff worked out with react-json-schema.

Besides the duplicate initInteractive that Kirk found. The only other thing I found was that I can't edit all of the interactives because they are too tall.  Perhaps we can make them scroll if the outer interactive is not able to get any taller?

package.json
```diff
@@ -107,6 +107,9 @@
     "webpack-dev-server": "^3.10.3"
   },
   "dependencies": {
+    "@fortawesome/fontawesome-svg-core": "^1.2.28",
+    "@fortawesome/free-solid-svg-icons": "^5.13.0",
+    "@fortawesome/react-fontawesome": "^0.1.9",
```
kswenson, line 112: There may be a larger plan here of which I'm unaware, but it looks like you're adding these several libraries as a means of getting access to a single fontawesome lock icon, which seems like a bit of overkill. What we did for the `slate-editor` work (and I think @knowuh did for some of the CLUE teacher dashboard work) was to download the SVG (e.g. this [fontawesome lock icon](https://www.iconfinder.com/icons/1608769/lock_icon)) and then convert it to a React component using the [SVGR](https://react-svgr.com/playground/) converter which results in a simple React component like this:
```
import * as React from "react"

function SvgComponent(props) {
  return (
    <svg height={1792} viewBox="0 0 1792 1792" width={1792} {...props}>
      <path d="M640 768h512V576q0-106-75-181t-181-75-181 75-75 181v192zm832 96v576q0 40-28 68t-68 28H416q-40 0-68-28t-28-68V864q0-40 28-68t68-28h32V576q0-184 132-316t316-132 316 132 132 316v192h32q40 0 68 28t28 68z" />
    </svg>
  )
}

export default SvgComponent
```
You can then add types for use with TypeScript and props (e.g. `fill`) and you end up with a component for each icon you need without the overhead of any additional dependencies. Just another option to consider.

pjanik: Right, I've been using SVGR each time Micheal is providing custom icons. It's nice then, but quite often I need to manually adjust SVG, e.g. viewbox or some little details. Not much work, but still. I'd prefer if we could pick one icon library for life. 😉 

Here I was assuming that these libraries will be only temporary - we don't have any design yet. But at some point we might end up with custom icons too. react-fontawesome looked good, as it was super simple to use and they're promising in readme that the way I imported lock icon, would import only this one SVG. So, I hoped it won't increase build size.

But since you mentioned that, I did a little test. Adding this single lock icon increases bundle size of all the interactives from 400KB to 1.07MB...  So, given that we'll have many of them on one LARA page, I think it's definitely worth refactoring for SVGR. Thanks.

PR #9, opened 2020-05-20
Add useRequiredQuestion hook and use it in all the question types

pjanik:
This hook should ensure that behavior is consistent across all the interactives. Also, if interactive ties to use it, but it doesn't implement `submitted` and `required` properties in its states, TypeScript will detect that and won't compile the code.

Theoretically, I could try to provide some wrapper even higher, but this implementation gives interactives flexibility on where to put a button and locked info. For example, Leslie wanted to have submit button on the right side of the question, while typical LARA questions have it on the left - so it's possible here pretty easily.

### DIFF
--- a/src/fill-in-the-blank/components/runtime.tsx
+++ b/src/fill-in-the-blank/components/runtime.tsx
@@ -1,14 +1,16 @@
 import React from "react";
 import { IAuthoredState, IBlankDef } from "./authoring";
 import css from "./runtime.scss";
+import { useRequiredQuestion } from "../../shared/hooks/use-required-question";
 
-interface IFilledBlank {
+export interface IFilledBlank {
   id: string;
   response: string;
 }
 
-interface IInteractiveState {
+export interface IInteractiveState {
   blanks: IFilledBlank[];
+  submitted?: boolean;
 }
 
 interface IProps {
@@ -16,6 +18,7 @@ interface IProps {
   interactiveState?: IInteractiveState;
   setInteractiveState?: (state: IInteractiveState) => void;
   report?: boolean;
+  setNavigation?: (enableForwardNav: boolean, message: string) => void;
 }
 
 export const insertInputs = (prompt: string, blanks: IBlankDef[], userResponses: IFilledBlank[]) => {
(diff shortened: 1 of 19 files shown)


src/shared/hooks/use-required-question.tsx
```diff
@@ -0,0 +1,44 @@
...
+      setInteractiveState(Object.assign({}, interactiveState, { submitted: true }));
+    }
+  };
+
+  useEffect(() => {
+    if (authoredState.required && setNavigation) {
+      const forwardNavEnabled = !!interactiveState?.submitted;
+      setNavigation(forwardNavEnabled, forwardNavEnabled ? "" : "Please submit an answer first.");
+    }
+  }, [authoredState, interactiveState, authoredState]);
```
scytacki, line 28: authoredState is repeated here

src/fill-in-the-blank/components/runtime.tsx
```diff
@@ -46,7 +49,10 @@ export const insertInputs = (prompt: string, blanks: IBlankDef[], userResponses:
   return result;
 };
 
-export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, setInteractiveState, report }) => {
+export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, setInteractiveState, setNavigation, report }) => {
+  const submitEnabled = (interactiveState?.blanks || []).length > 0;
```
scytacki, line 53: This might be more clear if the variable name and the one passed to useRequiredQuestion was `isAnswered` or `hasAnswer`.  I was confused because there are many things that can change the submitButton, so using submitEnabled seemed to conflict with the other logic.

pjanik:
I fixed this bug/typo and updated naming, thanks.

> It does seem like there ought to be a way to DRY up the App components in each of the question types. They seem to be mostly the same, but I'm not sure how to do that with

I agree. At the same time, there's not much logic there. Mostly about importing files and hooks. And they look identical, as we have very similar file structure for each interactive. So, at least we're mostly repeating some mechanical things, not the logic (which is mostly extracted). We could create a `BaseApp` which would accept `Runtime` and `Authoring` components and simplify things a bit. We could do the same with `BaseAuthoring` I think, as so far they're all very similar. Let's see how it goes with new components, I can do it next time I'm back in this repository.

PR #10, opened 2020-05-21
Fix height calculations

pjanik:
There were a couple of issues related to height calculations.

It's not enough to detect top-level component re-render. The scaffolded question authoring has its own internal state that changes height (and authoredState is not updated). This wasn't detected correctly by useAuthoHeight hook. I was thinking about a few different patterns and ended up using an optional context. Scaffolded question and its child components are an example of how to use it.

CSS transitions of height-related properties were also causing problems. Height recalculation might be triggered too early. That's where there are new `setTimeout` calls in `recalcHeight`.

It's all not too bad, but actually I'm wondering about an alternative solution. We could just trigger `recalcHeight` every 100ms - non-stop. These calculations take almost no time (console.time shows ~0.01ms). Then, it'd be enough to use this hook in the top-level component and forget about it. No need to worry if some child is changing its height too. I didn't do it, as I'm 100% sure what would happen on a slow machine with 10-20 interactives calling this function all the time. Or if there's some edge case where checking height is forcing the browser to do some internal work (layout, repaint, etc.). But I think that if we have some problems with height calculations in the future, this might be a bulletproof solution.

Finally, scrollHeight works better than offsetHeight (`e.g. when we have some elements with `float:right` css props). Generally it seems to make more sense, as scrollHeight describes necessary height of the container to avoid scrollbars.

### DIFF
--- a/src/shared/hooks/use-auto-height.ts
+++ b/src/shared/hooks/use-auto-height.ts
@@ -14,7 +14,7 @@ export const useAutoHeight = (config: IConfig) => {
   const currentHeight = useRef(0);
 
   const recalcHeight = (last = false) => {
-    const height = container.current?.offsetHeight;
+    const height = container.current?.scrollHeight;
     if (height && height > 0 && height !== currentHeight.current) {
       setHeight(height);
       currentHeight.current = height;
--- a/src/shared/hooks/use-auto-height.ts
+++ b/src/shared/hooks/use-auto-height.ts
@@ -1,4 +1,8 @@
-import { RefObject, useEffect } from "react";
+import React, { RefObject, useEffect, useRef } from "react";
+
+export const AutoHeightContext = React.createContext({
+  notifyHeightUpdated: () => { /* noop */ }
+});
 
 interface IConfig {
   container: RefObject<HTMLDivElement>;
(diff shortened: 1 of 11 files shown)


pjanik:
Generally, the idea is that the top-level component need to check its height from time to time and notify parent (LARA). But sometimes height change comes from the child components that get shorter or taller.

To avoid passing props all the way down in the component hierarchy, I've added a context with notify function.

The second option would be to remove all that code and just let the parent check its current height every 50/100/200/500ms all the time (e.g. using `setInterval`). Then, developers and child components don't need to care about anything related to height.

scytacki:
The browser be automatically recalculating the size of the element when its children change. So this feels like we are duplicating its existing functionality here. 
Can we used this ResizeObserver instead: 
https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver

kswenson:
@scytacki ResizeObserver has fairly limited browser availability, but there are libraries like those I linked above that offer similar functionality across a broader range of browsers.

scytacki:
This page compares several of these resize monitoring options: https://justineo.github.io/resize-detector/. Probably biased in some way though. But the comparison seems objective enough.

I do think it is worth prioritizing this now. Resizing is hard, and it makes sense to get right as best as we can. Which means it should be fast and require as little extra code in each component as possible.

pjanik:
Thanks! Actually, I didn't expect to see so many libraries and even native API doing this. I should have checked first. I ended up using native `ResizeObserver` + https://www.npmjs.com/package/resize-observer-polyfill as a backup. It was simple and native api seems always a bit safer / more efficient in the long-term. It seems to work well.

scytacki:
Looks great!.

src/shared/hooks/use-auto-height.ts
```diff
@@ -8,22 +9,25 @@ interface IConfig {
...
-    // only once, as useEffect has empty dependencies list.
-    const resizeHandler = () => calcHeight();
-    window.addEventListener("resize", resizeHandler);
-    return () => window.removeEventListener("resize", resizeHandler);
-  }, []);
-}
+    // TypeScript doesn't seem to have types of the native ResizeObserver yet. Use types coming from polyfill.
+    const NativeResizeObserver = (window as any).ResizeObserver as new(callback: ResizeObserverCallback) => ResizeObserver;
+
+    const observer = new (NativeResizeObserver || ResizeObserver)(entries => {
```
kswenson, line 16: I believe this is unnecessary, as the `ResizeObserver` returned by the module should be the [native ResizeObserver](https://github.com/que-etc/resize-observer-polyfill/blob/4a148452494a155656e4d04b9732b5914896b2e0/src/index.js#L6-L8) if it's available.

pjanik: Nice, I was wondering if that's the case or not. Although the docs suggest that it's a "ponyfill":
https://github.com/sindresorhus/ponyfill
And there's a statement:
> Ponyfills should never use the native API, even if available, as it might have slightly different behavior between environments, which can cause bugs.

So, I thought it's just safer to do the check ourselves anyway.

kswenson: I followed that same thread and concluded that the recommended use is not actually a ponyfill according to Sindre's definition, but it's also not a polyfill since it uses the native implementation if it's present but doesn't patch it in if it's not. I guess we need a new word. Maybe it's a Posyfill, because it poses as the native implementation if it's not present. 😉

PR #13, opened 2020-05-25
Hints and refactoring

pjanik:
[#172821069]

This PR adds support of "hint" Interactive API message: 
https://github.com/concord-consortium/lara/pull/530

Also, adds `BaseQuestionApp` and `BaseAuthoring` to reduce boilerplate that was required for each question type. We mentioned that a few times in previous PRs. `BaseQuestionApp` provides set of opinionated features that probably every question in LARA should support (height, hint, required/submitted). There's a bit of generic typing there, but regular interactive author won't have to mess with that too much. All the features are optional and based on the authored and interactive state. For example, "Submit" button will be displayed only if authored state has "required" property. Name of these properties can be easily guessed from interfaces or existing interactive examples.

`BaseAuthoring` is a wrapper around react-jsonschema-form authoring.

I've refactored only `OpenResponse` - I think it's better to get feedback before I refactor other interactives. But it should show how much less code we have now. Pretty much just types, JSON schema and Runtime.


src/open-response/components/runtime.tsx
```diff
@@ -41,17 +30,6 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
...
       </div>
-      {
-        authoredState.extraInstructions &&
-        <div className={css.extraInstructions}>{ authoredState.extraInstructions }</div>
-      }
-      {
-        !report &&
-        <div>
-          { submitButton }
-          { lockedInfo }
```
scytacki, line 52: We might need to do something more advanced here so the individual questions can control where the submitButton and lockedInfo go in their UI. I'm assuming that they get automatically added below the question with this change. I'm confident that we can figure out how to do that if and when it is needed.

src/shared/components/base-authoring.tsx
```diff
@@ -0,0 +1,47 @@
...
+  setAuthoredState?: (state: IAuthoredState) => void;
+  preprocessFormData?: (data: IAuthoredState) => IAuthoredState;
+  // react-jsonschema-form properties:
+  schema: JSONSchema6;
+  uiSchema: UiSchema;
+  // react-jsonschema-form additional fields.
+  fields?: { [name: string]: Field };
+}
+
+export const BaseAuthoring = <IAuthoredState,>({ authoredState, setAuthoredState, preprocessFormData, schema, uiSchema, fields }: IBaseAuthoringProps<IAuthoredState>) => {
```
kswenson, line 19:
```suggestion
export const BaseAuthoring = <IAuthoredState>({ authoredState, setAuthoredState, preprocessFormData, schema, uiSchema, fields }: IBaseAuthoringProps<IAuthoredState>) => {
```

src/shared/components/base-authoring.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const onChange = (event: IChangeEvent) => {
+    let formData = event.formData;
+    if (preprocessFormData) {
+      formData = preprocessFormData(formData);
+    }
+    // Immediately save the data.
+    if (setAuthoredState) {
+      setAuthoredState(formData);
```
kswenson, line 27:
```suggestion
      setAuthoredState?.(formData);
```

src/shared/components/base-question-app.tsx
```diff
@@ -0,0 +1,85 @@
...
+  useHint({ authoredState, setHint });
+  const { submitButton, lockedInfo } = useRequiredQuestion({ authoredState, interactiveState, setInteractiveState, setNavigation, isAnswered: isAnswered(interactiveState) });
+
+  const report = mode === "report";
+  const authoring = mode === "authoring";
+  const runtime = mode === "runtime";
+
+  let AuthoringComp = null;
+  if (authoring) {
+    if (Authoring) {
```
kswenson, line 58: This use of local variable names that differ only in case confused me initially. Consider `reportMode`, `authoringMode`, `runtimeMode` or some other such renaming to make the distinction more readily apparent.

src/shared/components/base-question-app.tsx
```diff
@@ -0,0 +1,85 @@
...
+      { AuthoringComp }
+      {
+        (runtime || report) && authoredState &&
+        <div className={css.runtime}>
+          <Runtime authoredState={authoredState} interactiveState={interactiveState} setInteractiveState={setInteractiveState} report={report} />
+          {
+            !report &&
+            <div>
+            { submitButton }
+            { lockedInfo }
```
kswenson, line 77: Indentation?

src/shared/components/base-question-app.tsx
```diff
@@ -0,0 +1,85 @@
...
+    }
+    if (!Authoring && baseAuthoringProps) {
+      AuthoringComp = <BaseAuthoring {...baseAuthoringProps} authoredState={authoredState} setAuthoredState={setAuthoredState} />;
+    }
+  }
+
+  return (
+    <div ref={container}>
+      { AuthoringComp }
+      {
```
scytacki, line 69: This was a little confusing to me since it seems like it is possible to show both the AuthoringComp and the Runtime component at the same time. I realized that isn't possible because the AuthComp will be null if the mode is authoring or runtime. 

Perhaps it is worth putting this in a named method or self invoking method and use switch case like:
```
switch(mode){
case "authoring": 
  if(Authoring)....
case "runtime":
  return <div ><Runtime authoredState={authoredState} interactiveState={interactiveState} setInteractiveState={setInteractiveState} report={false} />
<div>{submitButton} {lockedInfo}</div><div/>
case "report":
  return <Runtime authoredState={authoredState} interactiveState={interactiveState} setInteractiveState={setInteractiveState} report={true} />
default:
  return "Loading...";
```
It is a little redundant but it seems more clear. To reduce the redundancy the runtime and report body could be pulled out to a function that takes a report variable so then it'd just be:
```
switch(mode){
case "authoring": 
  if(Authoring)....
case "runtime":
  return renderRuntime({report:false})
case "report":
  return renderRuntime({report:true})
default:
  return "Loading...";
```

src/shared/hooks/use-lara-interactive-api.ts
```diff
@@ -35,15 +35,19 @@ export const useLARAInteractiveAPI = <AS, IS>(config: IConfig) => {
...
+  };
 
   const handleSetHeight = (height: number) => {
     phone.current?.post("height", height);
-  }
+  };
+
+  const handleSetHint = (hint: string) => {
+    phone.current?.post("hint", hint);
+  };
```
scytacki, line 46: This is something we'll have to work out in the lara interactive client, but it seems possible that the hint message could be posted to LARA before supportedFeatures is sent to LARA. I guess that is OK, but we should clarify whether that is allowed or not.

cypress.json
```diff
@@ -1,5 +1,5 @@
 {
-  "baseUrl": "http://localhost:8080",
+  "baseUrl": "http://localhost:8081",
```
scytacki, line 2: It looks like this change broke travis's running of the cypress tests.

src/shared/components/base-authoring.tsx
pjanik: This comma is actually necessary. Otherwise, TS is getting confused - it doesn't know if `<IAuthoredState>` is a generic type or JSX syntax. ;) If there are two types provided, it's already fine.

src/open-response/components/runtime.tsx
pjanik: Right. To make it work with ScaffoldedQuestion, I thought I'd just let interactive provide a set of buttons that will be added to the bottom row + let it specify CSS class name (so it can be moved to the right).

If things get more complex, I think there are two approaches:
  - Interactive doesn't have to use this BaseQuestionApp. It can implement its own, it's not super complicated. I think we can a few different BaseApps if a few interactives share common patterns rather than trying to make one super configurable helper.
  - Hooks cannot be conditional (rules of hooks), but we can pass something like `isEnabled` to each of them. That way, we could turn on and off various features. So, an interactive could turn off `useRequiredQuestion` and then use this hook itself in the runtime (well, in fact we don't even have to turn it off, it won't hurt if it's in both places). This will work when we `useLARAInteractiveAPI` uses global iframe-phone and can be requested multiple times. Then, all the hooks will also become much simpler, as each hook will use `useLARAInteractiveAPI`.

src/shared/components/base-question-app.tsx
```diff
@@ -0,0 +1,85 @@
...
+export const BaseQuestionApp = <IAuthoredState extends IBaseQuestionAuthoredState, IInteractiveState extends IBaseQuestionInteractiveState>(
+  { Authoring, baseAuthoringProps, Runtime, isAnswered }: IProps<IAuthoredState, IInteractiveState>
+) => {
+  const container = useRef<HTMLDivElement>(null);
+  const { mode, authoredState, interactiveState, setInteractiveState, setAuthoredState, setHeight, setHint, setNavigation } = useLARAInteractiveAPI<IAuthoredState, IInteractiveState>({
+    interactiveState: true,
+    authoredState: true,
+  });
+  useAutoHeight({ container, setHeight });
+  useHint({ authoredState, setHint });
```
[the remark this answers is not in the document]
pjanik: I mentioned that in another comment, but we can pass `isEnabled` to each hook and disable it "inside". 

Most of the features/hooks are optional and depend on authored and interactive state - e.g. hint or required question. So we don't really have to turn them off. If authored state doesn't specify `required` or `hint`, they will just do nothing. 

I think the only thing that we might need to turn off is `useAutoHeight`.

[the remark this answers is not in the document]
scytacki: Sounds good. Yes it seems like useAutoHeight will be the only one that will need to be explicitly disabled.

src/open-response/components/runtime.tsx
scytacki: Yup sounds good.

pjanik:
Thanks @scytacki and @kswenson. I pushed a few more commits to address most of the feedback. Or I left some comments with explanations. So, I think every comment should be addressed now.

> Right. To make it work with ScaffoldedQuestion, I thought I'd just let interactive provide a set of buttons that will be added to the bottom row + let it specify CSS class name (so it can be moved to the right).

I was trying this approach, but it's not great that way. Generally, being able to use LARA API hook multiple times should simplify things a lot. And I found another way to deal with submit button issue, but I'll open a separate PR for that.

PR #14, opened 2020-05-27
Hints and refactoring 2

pjanik:
This PR continues refactoring started in: https://github.com/concord-consortium/question-interactives/pull/13

All the interactives has been refactored to use `BaseQuestionApp`, so they all support hints correctly now. New options have been added to `BaseQuestionApp`, so interactive might disable auto height or rendering of the submit button.

`ScaffoldedQuestion` renders the submit button and lock info itself. To do it, I removed the submit button rendering from `useRequiredQuestion` hook and added it separately as a small component.

Once we have `useLARAInteractiveAPI` hook that can be used multiple times (based on global/module state), it should simplify other hooks.


src/shared/components/base-authoring.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+  }
+} as JSONSchema6;
+
+const authoredState = {
+  version: 1,
+  testValue: "123"
+};
+
+describe("BaseAuthoring", () => {
+  it("renders react-jsonschema-form and passes there authoredState", () => {
```
kswenson, line 32:
```suggestion
  it("renders react-jsonschema-form and passes the authoredState", () => {
```

src/shared/components/base-authoring.test.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const formEl = wrapper.find(Form);
+    expect(formEl.length).toEqual(1);
+    expect(formEl.props().formData).toEqual(authoredState);
+    expect(formEl.props().schema).toEqual(schema);
+    expect(formEl.props().uiSchema).toEqual(uiSchema);
+    expect(formEl.props().fields).toEqual(fields);
+    expect(formEl.props().validate).toEqual(validate);
+  });
+
+  it("calls setAuthoredState on form change and uses preprocess form data function", () => {
```
kswenson, line 56:
```suggestion
  it("calls setAuthoredState on form change and uses preprocessFormData function", () => {
```

src/shared/components/base-authoring.tsx
```diff
@@ -1,40 +1,56 @@
...
+
   const onChange = (event: IChangeEvent) => {
     let formData = event.formData;
     if (preprocessFormData) {
       formData = preprocessFormData(formData);
     }
     // Immediately save the data.
     setAuthoredState?.(formData);
+    if (validate) {
+      triggerDelayedValidation();
```
kswenson, line 33: Could also use (here and below):
```suggestion
      validate && triggerDelayedValidation();
```

pjanik: I applied other suggestions, thanks! I might skip this one for now though, as I think in most of the code I use explicit `if` statements in similar cases (except for JSX where it's hard). So, just to be consistent. Personally, I like this type of syntax, but I stopped using it over time as in some cases people mention that it's not super readable. It is to me, but I understand the other point too. ;)

PR #15, opened 2020-05-29
Refactor codebase to use new LARA Interactive Client

pjanik:
[#173082951]


src/shared/components/submit-button.tsx
```diff
@@ -1,17 +1,18 @@
...
-  authoredState: { required?: boolean } | undefined;
-  interactiveState: { submitted?: boolean } | undefined;
-  setInteractiveState: ((interactiveState: {submitted?: boolean}) => void) | undefined;
   isAnswered: boolean;
 }
 
 // This hook can be used by any interactive that defines `required` property in its authored state and
 // `submitted` property in its interactive state (student state).
-export const SubmitButton: React.FC<IProps> = ({ authoredState, interactiveState, setInteractiveState, isAnswered }) => {
+export const SubmitButton: React.FC<IProps> = ({ isAnswered }) => {
```
scytacki, line 12: I'm not sure it is worth it, but it'd be cool if the SubmitButton was passed the authored state type and the interactiveState type.  Then it could cause a type error if these types don't have `required?` and `submitted?`. 

It looks like this is possible based on this post: https://mariusschulz.com/blog/passing-generics-to-jsx-elements-in-typescript

src/shared/hooks/use-required-question.tsx
```diff
@@ -1,18 +1,17 @@
...
-  authoredState: { required?: boolean } | undefined;
-  interactiveState: { submitted?: boolean } | undefined;
-  setNavigation: ((enableForwardNav: boolean, message: string) => void) | undefined;
-}
+import { setNavigation, useAuthoredState, useInteractiveState, useInitMessage } from "@concord-consortium/lara-interactive-api";
 
 // This hook can be used by any interactive that defines `required` property in its authored state and
 // `submitted` property in its interactive state (student state).
-export const useRequiredQuestion = ({ authoredState, interactiveState, setNavigation }: IConfig) => {
+export const useRequiredQuestion = () => {
```
scytacki, line 6: If the submit button took the interactiveState type and authoredState type, they could be passed into this hook as generics as well to provide some type checking to verify what is described in the comment on line 4 and 5.

scytacki:
This looks really good to me.

src/shared/components/submit-button.tsx
pjanik: My first implementation was like this. SubmitButton wasn't extracted back then, but basically useRequiredQuestion would complain if there were no `required?/submitted?` props. But then I changed it to make it less strict, since some interactives/question might not want to support the submitt button (or hint or other features that we add later).

Now, they can still use the same `BaseQuestionApp` without any configuration. `requiored?` ensures that required prop might be defined, but it doesn't have to. If it's not, it's all good and the feature is disabled.

This typing actually more helps to write `useRequiredQuestion` or `SubmitButton`, as we're adding some (loose) assumptions about passed in states.

I can change it to be more strict again, but then we would need to add more configuration to `BaseQuestionApp` and add `disabled` props to most of the hooks. Or don't add `disabled` and assume that `BaseQuestionApp` is forcing developer to support these features (since most questions actually should support submit button, hint, etc.).

pjanik:
@scytacki, I was trying a similar approach before and tried it again now.
Basically, this typing would make sense if we're using:
```
submitted: boolean | undefined;
required: boolean | undefined;
```
instead of:
```
submitted?: boolean;
required?: boolean;
```

These definitions seem almost the same, but they are not. The first one is more strict. It requires interfaces to define these properties. The second one defines optional params, so the passed in interfaces might define these properties, but they don't have to.

So, we have two choices: 
1. `BaseQuestionApp` ensures that developer will always implement all the features used by it - submitted/required question, hint, and whatever else we add in the future. It's nice, as we have some required interface for every question. So, developer will be "forced" to provide expected authored and interactive state formats. And then the extended generic typing that you described in comments works well, as they fit this enforcement. It's an easy change, I'd need to change a few `prop?: boolean` to `prop: boolean | undefined`.
2. `BaseQuestionApp` provides a base for many different interactives and some of them don't need to support required question or hints (or any new features that we add in the future). This is the current implementation with many properties using `?`. In such case, this extended typing doesn't change anything, as everything is optional and there are no errors anyway.

Any preferences? As I said, we currently have 2. I think 1 is nice, but it's just a bit different idea. If someone adds non-question interactive, he'd have to create a new base (or just custom app).

scytacki:
I think option 2 is sounds better. We definitely have interactives that shouldn't support required questions. Thanks for looking into this so thoroughly.

pjanik:
Okay. Just to be 100% clear, in option 2, we could have `BaseQuestionApp` and separate `BaseApp` (for regular, non-question interactives). So, the main question is whether all the questions should be required to implement these standard features like required or hint. If so, then I think it could actually make sense. As a developer using `BaseQuestionApp` would be automatically "forced" to add all the features that we think every question should have. 

If something is not considered to be a "question", it could use `BaseApp` instead and not be forced to implement required feature (usually it's just authored and interactive state props).

pjanik:
OK, this is ready for re-review and/or to be merged. I've used the very last version of lara-interactive-api NPM package, did some cleanup and added support to functional updates of interactive state in all the questions.

Actually, I'm pretty much forcing all the interactives to use functional updates. A bit opinionated, but since we discussed that, it's clear they're safer. And the interactive state can be sometimes used in a few places (like in a submit button that uses it independently). Still, I don't think it's very likely some data would get lost, but since it's a student data, I guess it's better to be safe.

pjanik:
There's one Cypress test failing sometimes, it's unrelated to these changes. I'll try to fix it tomorrow.

pjanik:
> I'm curious why the BaseApp passes the interactiveState and setInteractiveState through the props to the runtime. Couldn't the runtime just use the useInteractiveState hook directly?

It started like that, as old Interactive API hook could be used only once, so it was in the top-level component. Now, it's different and you're right, each Runtime could use `useInteractiveState` itself. I thought about it but didn't change, as I have no strong opinion here. Maybe props make it even a bit simpler, as you don't even have to import the hook and think how to use it. But it's simple. And you're right about props. 

I was thinking about `BaseQuestionApp` as a base for questions specifically (not all the interactives). So these props make it a bit more obvious that questions should handle interactive state. But I think it's a tiny difference and if for whatever reason someone feels like it's better to use hook directly, I think it's totally fine to refactor.

PR #17, opened 2020-06-05
Image interactive

dstrawberrygirl:
This PR addresses the main display and authoring modes for the interactive.
Addressing height (fitting the image to the be completely visible on screen) will be addressed later / after discussion with other devs to figure out the best way to make it happen.


src/image/components/runtime.test.tsx
```diff
@@ -0,0 +1,38 @@
...
+    const wrapper = shallow(<Runtime authoredState={authoredState} />);
+    expect(wrapper.text()).toEqual(expect.stringContaining(authoredState.caption));
+  });
+});
+describe("report mode", () => {
+  it("renders caption and image", () => {
+    const wrapper = shallow(<Runtime authoredState={authoredState} report={true} />);
+    expect(wrapper.text()).toEqual(expect.stringContaining(authoredState.caption));
+  });
+});
```
pjanik, line 38: This file doesn't seem finished :wink:

cypress/integration/image.test.js
```diff
@@ -0,0 +1,90 @@
...
+      cy.getIframeBody().find("#root_caption").should("include.text", authoredStateSample.caption);
+      cy.getIframeBody().find("#root_credit").should("include.text", authoredStateSample.credit);
+      cy.getIframeBody().find("#root_creditLink").should("have.value", authoredStateSample.creditLink);
+      cy.getIframeBody().find("#root_creditLinkDisplayText").should("have.value", authoredStateSample.creditLinkDisplayText);
+      cy.getIframeBody().find("#root_allowLightbox").should("be.checked");
+
+    });
+  });
+
+});
```
pjanik, line 90: I know it's pretty repetitive, but we have already set of cypress tests for other interactives. They could be just copy-pasted and adjusted. This seems to do it partially, but some parts are committed (interactive state saving, authoring updates). It'd be nice if we keep that consistent. I've been refactoring this repo quite a lot, and both cypress and unit tests were actually useful to do it more confidently.

scytacki:
@pjanik the displaying of the bar above the interactive depends on whether the authoring system has enabled saveState on the interactive. So in theory this image interactive can be added to LARA with saveState disabled.  I'm planning to do that with the video interactive, so there will be two versions: one that saves state (and has the bar), and one that doesn't save state (no bar). That way authors can choose.

However @dstrawberrygirl I do agree that for the image interactive, just saving whether the user viewed the interactive isn't worthwhile. If we did that, then every interactive should save this viewed property.  If we want to support that I think it would be better to add it into LARA core instead of the individual interactives. Now that Piotr added a new BaseApp class to the repo, I think you can easily drop the interactiveState from the image interactive.

dstrawberrygirl:
Thanks @pjanik and @scytacki - will remove the state and update tests.

src/video-player/components/runtime.test.tsx
```diff
@@ -36,4 +36,12 @@ describe("Runtime", () => {
     const wrapper = shallow(<Runtime authoredState={authoredState} interactiveState={interactiveState} />);
     expect(wrapper.text()).toEqual(expect.stringContaining(interactiveState.lastViewedTimestamp.toString()));
   });
+
+  it("parses aspect ratio", () => {
+    const wrapper = shallow(<Runtime authoredState={authoredState} interactiveState={interactiveState} />);
```
pjanik, line 41:
```suggestion
```

This line doesn't seem to be necessary.

PR #18, opened 2020-06-10
Use Authoring and Runtime metadata

pjanik:
Fill-in-the-blank needs to be converted, I'll do it in the next PR.

### DIFF
--- a/src/multiple-choice/components/app.tsx
+++ b/src/multiple-choice/components/app.tsx
@@ -3,40 +3,22 @@ import { JSONSchema6 } from "json-schema";
 import { BaseQuestionApp } from "../../shared/components/base-question-app";
 import { Runtime } from "./runtime";
 import { v4 as uuidv4 } from "uuid";
+import {
+  IAuthoringMultipleChoiceChoiceMetadata, IAuthoringMultipleChoiceMetadata, IRuntimeMultipleChoiceMetadata,
+} from "@concord-consortium/lara-interactive-api";
 
 // Note that TS interfaces should match JSON schema. Currently there's no way to generate one from the other.
 // TS interfaces are not available in runtime in contrast to JSON schema.
 
-export interface IChoice {
-  id: string;
-  content: string;
-  correct?: boolean;
-}
+export interface IChoice extends IAuthoringMultipleChoiceChoiceMetadata {}
 
-// Note that format of this metadata is pretty strict. It needs to match LARA and report-service expectations.
-// It can be moved to lara-interactive-api package.
-export interface IAuthoringMetadata {
-  questionType: "multiple_choice";
-  prompt?: string;
-  required?: boolean;
-  choices: IChoice[];
-}
-
-export interface IAuthoredState extends IAuthoringMetadata {
+export interface IAuthoredState extends IAuthoringMultipleChoiceMetadata {
   version: number;
   hint?: string;
   multipleAnswers?: boolean;
 }
 
-// Note that format of this metadata is pretty strict. It needs to match LARA and report-service expectations.
-// It can be moved to lara-interactive-api package.
-export interface IRuntimeMetadata {
-  answerType: "multiple_choice_answer",
-  selectedChoiceIds: string[];
-  submitted?: boolean;
-}
-
-export interface IInteractiveState extends  IRuntimeMetadata{}
+export interface IInteractiveState extends IRuntimeMultipleChoiceMetadata {}
 
 export const baseAuthoringProps = {
   schema: {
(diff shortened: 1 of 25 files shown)


src/multiple-choice/components/app.tsx
```diff
@@ -3,29 +3,22 @@ import { JSONSchema6 } from "json-schema";
...
 
 // Note that TS interfaces should match JSON schema. Currently there's no way to generate one from the other.
 // TS interfaces are not available in runtime in contrast to JSON schema.
 
-export interface IChoice {
-  id: string;
-  content: string;
-  correct?: boolean;
-}
+export interface IChoice extends IAuthoringMultipleChoiceChoiceMetadata {}
```
scytacki, line 13: Looks like this interface isn't used by IAuthoredState anymore.
But there is a need to add addition properties to this IChoice interface.
So perhaps IAuthoringMultipeChoiceMetadata needs to support passing a type for the choices array.

pjanik: > But there is a need to add addition properties to this IChoice interface.

You can do it now. Just use custom `choice` type in authored state (option 1). Or we can add generics (option 2).

1. Extend:
```
// No changes in Interactive API Client types necessary.

// Changes in question-interactives:

export interface IChoice extends IAuthoringMultipleChoiceChoiceMetadata {
  feedbackText: string;
}

export interface IAuthoredState extends IAuthoringMultipleChoiceMetadata {
  version: number;
  hint?: string;
  multipleAnswers?: boolean;
  choices: IChoice[]; // use custom IChoice - TS will guarantee that IChoice implements necessary fields (id, content, correct)
}
```

2. Generics:
```
// Changes in Interactive API Client types:

export interface IAuthoringMultipleChoiceChoiceMetadata {
  id: ChoiceId;
  content?: string;
  correct?: boolean;
}

export interface IAuthoringMultipleChoiceMetadata<IChoiceMetadata extends IAuthoringMultipleChoiceChoiceMetadata> extends IAuthoringMetadataBase {
  questionType: "multiple_choice";
  choices: IChoiceMetadata[];
}

// Changes in question-interactives:

export interface IChoice extends IAuthoringMultipleChoiceChoiceMetadata {
  feedbackText: string;
}

export interface IAuthoredState extends IAuthoringMultipleChoiceMetadata<IChoice> {
  version: number;
  hint?: string;
  multipleAnswers?: boolean;
}
```

I have no strong opinion here. Generic is more clear that you can use custom choice, but for a less TS-experienced dev looks scarier. Generally, the code looks more complicated. Option 1 hides a bit fact that you can provide custom fields (hence your question), but types are simpler. 

In terms of type checking, both options are exactly the same. The developer would be forced to provide choice that implements `id` and optionally `content` and `correct` fields in both cases.

pjanik: For now I assumed that option 1 is fine. I made sure that IChoice interface is used in one of the new commits.

pjanik:
> As I mentioned in the LARA PR it would also be nice if client type checking could enforce that if the IAuthoredState had a questionType of multiple choice, then it also included the multiple choice fields and that the IInteractiveState has the correct multiple choice answer fields. But I don't know how hard that would be.

For now, I just added this:
https://github.com/concord-consortium/question-interactives/pull/18/commits/2e9ac7ef2ba6979ce9776d19de5e0a0a6584321d
That forces interactives to use metadata, but it doesn't ensure that combination of authoring and runtime make sense. :wink: I didn't have an idea how to do it.

I've also converted all remaining interactives to use metadata (fill in the blank and video).

src/multiple-choice/components/app.tsx
sfentress: @pjanik My work merged in fine, but I do have a question, which is probably just me missing some context:

Why do we have *any* properties in these interfaces that aren't specified in the LARA Interactive API?

For example, with this current PR, the multiple-choice `IAuthoredState` still needs to add the properties `hint?` and `multipleAnswers?`, and will eventually define `layout?`, `givePredictionFeedback?`, `predictionFeedback?` etc as defined by LARA's current schema. Is there a reason not to have these all come from the LARA API types?

scytacki: The reason is because LARA doesn't need to care about these properties. This same api should be used by interactives like CBio and GeoCode. Those have many properties in their state that LARA doesn't care about either. The hint property is just a place to store information that will be sent to LARA using the hint api. multipleAnswers is used by the interactive to change its UI (the reporting metadata handling of this is via multiple choice objects that have a correct flag on them), layout is just presentation info ....

So the idea is to only add properties to the LARA interactive client api which are required somewhere down the chain.  As Piotr pointed out somewhere else, LARA itself doesn't actually need to know about most (perhaps all) of these new properties, it is really the reports that need them. 

Originally I had designed this, so these 'metadata' properties were stored outside of the authored and interactive state which might make this more clear.  But doing that makes the implementation more complex. So currently these special properties are mixed in the root of the authored and interactive state.

PR #21, opened 2020-06-12
Add prediction feedback

sfentress:
This adds `predictionFeedback?` to the base question, and displays it in the Lock component if it exists.

Screen Shot 2020-06-12 at 1 03 09 PM

Note that the old LARA models had both `give_prediction_feedback: boolean` and `prediction_feedback: string`, both of which had to be set to see any feedback. It seemed simpler to condense into just `predictionFeedback: string`. We'll have to remember this when creating the migration, however.

### DIFF
--- a/cypress/integration/multiple-choice.test.js
+++ b/cypress/integration/multiple-choice.test.js
@@ -62,6 +62,71 @@ context("Test multiple-choice interactive", () => {
     });
   });
 
+  context("Submit button", () => {
+
+    it("renders a submit button if required: true", () => {
+      phonePost("initInteractive", {
+        mode: "runtime",
+        authoredState: {
+          prompt: "Test prompt",
+          multipleAnswers: false,
+          choices: [
+            {id: "id1", content: "choice A"},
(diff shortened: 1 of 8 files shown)


pjanik:
Looks great to me, thanks. 
One suggestion would be to try to use schema dependencies. JSONSchema and react-jsonschema-form seem to support that. Then we could show prediction feedback input only if the question is required (as otherwise this feedback is never shown, right?).

I think it might be enough to use basic conditional dependency if unchecked/false value is treated the same as "undefined":
https://react-jsonschema-form.readthedocs.io/en/latest/usage/dependencies/#conditional
Otherwise, dynamic should work (but the syntax is more complex):
https://react-jsonschema-form.readthedocs.io/en/latest/usage/dependencies/#dynamic

sfentress:
Thanks, I started looking at conditional dependencies on Friday, and was not able to get it to work easily, as it was hidden with `undefined` but visible with `false`. I expect it should be simple enough, though, so will keep looking.

PR #22, opened 2020-06-12
172868311 fix video snapshot

scytacki:
See this PT story for details: https://www.pivotaltracker.com/story/show/172868311


scytacki:
@kswenson what do you do when the types don't support a few feature of the library? 
I figure this happens often.

kswenson:
The short-term answer is to suppress the error, for which the `@ts-ignore` you've added here is a fine solution. You might consider adding an additional comment to the effect of "The continuing need for this line should be verified as new versions of the types become available." The goody-two-shoes answer is to submit a PR to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) to address the problem but that's obviously more work.

PR #23, opened 2020-06-15
Update some strings to follow UDL

pjanik:
https://www.pivotaltracker.com/story/show/173209029


pjanik:
Thanks, I wasn't even sure if the PR is worth opening :wink: Apparently, it was. FIxed, so I'm merging it.

PR #25, opened 2020-06-15
Enable Shutterbug on all interactives

dstrawberrygirl:
I moved this to a hook and it seems to work well on the image and video interactives.


src/shared/hooks/use-shutterbug.ts
```diff
@@ -0,0 +1,12 @@
+import { useEffect } from "react";
+import Shutterbug from "shutterbug";
+
+// This hook can be used by any interactive that requires snapshots.
+export const useShutterbug = () => {
+  useEffect(() => {
+    Shutterbug.enable("#app");
```
kswenson, line 7: If you pass the selector as an argument (perhaps defaulting to `#app`) the hook would be usable in more contexts.

pjanik:
I think it wouldn't work now. `css.runtime` is a class name, not a selector that Shutterbug expects. That's why I mentioned `"." + css.runtime` with this strange `"."` at the beginning. ;)

dstrawberrygirl:
@pjanik it was working when I tested, but I think you're probably right so I just reworked it to pass in the "." and the selector and it works fine for both image and question.

PR #27, opened 2020-06-16
Fix authoring mode and runtime for images

dstrawberrygirl:
This changes the behavior of the Interactives relating to the `autoHeight` feature, enabling the use of interactives that require that their height is calculated from their aspect ratio.

- For authoring mode, we need to always use the `setHeight` call to resize the interactive to account for the dynamically-generated form fields.

- For runtime mode, this will allow the component to pass an `aspectRatio` to the LARA container via a `setSupportedFeatures` call, so an image component can resize a tall image to fit the window in responsive layout.

- For the preview mode runtime rendering, a fix has been implemented in https://github.com/concord-consortium/lara/pull/568 to ensure that part of the image is displayed at minimum.


scytacki:
One thing that is at least worth a comment in the code is:
Once the interactive has set its height it cannot 'unset' the height.
This might not always be the case, but it is today.

So this means that if someone enables auto height, they cannot successfully disable it later. I don't think this case is possible with the code in this PR. But since the PR is adding generic support to useAutoHeight and to BaseApp, it is worth at least adding a comment.

To be more strict, useAuthoHeight, could throw an exception if it ever switches to disabled after being enabled.

pjanik:
> To be more strict, useAuthoHeight, could throw an exception if it ever switches to disabled after being enabled.

I like the idea. Or we could add it to the client code to be even safer, as I think Doug just added similar height handling code in Vortex.

A little thing I'd consider would be splitting `disableAutoHeight` into `disableAuthoringAutoHeight` and `disableRuntimeAutoHeight` or just rename to `disableRuntimeAutoHeight`, so it's clear. But it doesn't seem we're going to use any custom authoring soon and why we would disable auto height there. So, not a big deal.

The last thing, not really related to this codebase, would be to extract setAspectRatio to a separate Interactive API call / method. That way `Runtime` wouldn't have to use `setSupportedFeatures` call that seems scarier and requires to send some other options there again (that are handled already by BaseQuestionApp). Plus, it'd be more similar to setHeight.

pjanik:
It looks good, sorry for late review, I didn't notice it's ready again.

scytacki:
@kswenson Could you review, fix conflicts and merge?

kswenson:
>@kswenson Could you review, fix conflicts and merge?

After merge it no longer seems to size the interactive correctly. Looking into it.

scytacki:
After experimenting with this here: 
https://github.com/concord-consortium/question-interactives/pull/38

I think we should not take the approach of this PR. 
We should back out all of the usage of the supportedFeatures.aspectRatio.

PR #28, opened 2020-06-17
Add Check Answer button and choice feedback

sfentress:
This adds the ability to author a check answer button, and to add optional custom feedback per choice.

The correct/incorrect default text, and behavior for multiple answers is the same as in current LARA.

This also adds placeholder text for pulldown options, to ensure student selects an answer before pressing Submit.


scytacki:
@dstrawberrygirl and perhaps @kswenson can you take a look at this one? We want to get this change in so Carolyn can start authoring and we are less likely to need to migrate the authored data later.

I looked at the schema for the authored data and it seems to be fine to me.

dstrawberrygirl:
This looks great to me - a lot of variables to consider but seems to work well. A very minor point, we seem to allow for the use case where multiple options could be correct but also have the ability to restrict students to choosing a single answer without giving an author any feedback on their design.

scytacki:
I'm going to merge this because it has passed 2 reviews. Feel free to merge without my approval.

PR #31, opened 2020-06-19
Add useStudentSetting hook, and use settings in ScaffoldedQuestion

pjanik:
[#173088322]

Related to: https://github.com/concord-consortium/glossary-plugin/pull/73

This `useStudentSetting` should be reusable in all components and shouldn't download FirebaseJWT if it's been already downloaded. I didn't wrap up tests today, but maybe it's still worth reviewing.


src/shared/hooks/use-student-settings.ts
```diff
@@ -0,0 +1,116 @@
+import * as firebase from "firebase";
+import "firebase/firestore";
+import { useEffect, useState } from "react";
+import { getFirebaseJWT } from "@concord-consortium/lara-interactive-api";
+import { IJwtResponse } from "@concord-consortium/lara-interactive-api";
```
kswenson, line 5: Could combine into one import.

scytacki:
Looks good to me. I'll wait to do a final review until there are tests.

pjanik:
@scytacki, I've added tests. I ended up splitting the file into use-student-settings and use-student-settings-helpers because of testing and Jest mocks. It can make sense anyway, as helpers are React-independent and include Firestore-based code mostly.

PR #32, opened 2020-06-25
Add logging

pjanik:
I've added `useBaseLogging` and used it in `BaseQuestionApp` and `BaseApp`. So, every interactive will log by default:
- "focus in", focus_target: ...
- "focus out"
- "answer saved", answer_text: ... IF interactive specifies `answerText`, sent on focus out event
This is what LARA used to log. Additionally, new interactives will also log:
- "answer submitted", answer_text: ... IF interactive specifies `answerText` and uses submit button

That's why I changed answerText of Scaoffolded Question to be more informative and added answerText to MC question.

Scaffolded question:
It doesn't use base logging, as it didn't work too well. There's a comment in the code explaining this - most of the input elements are inside iframe, so focus in/out events aren't detected. They can only come from button click, but then the event stream looks confusing. So, it forwards logs from subquestion and adds extra info like current level, subinteractive_url, subinteractive_type, subinteractive_sub_type. Plus, it uses default Submit button logging, and a hint button click is logged too.

Additional / custom events in other interactives:
- "video started", video_url: ..., percentage_viewed: ...
- "video stopped", video_url: ..., percentage_viewed: ...
- "image zoomed in", url: ...

There are Jest tests of the hook and Cypress test of Scaffolded Question logging.

### DIFF
--- a/cypress/integration/scaffolded-question.test.js
+++ b/cypress/integration/scaffolded-question.test.js
@@ -113,37 +113,38 @@ context("Test Scaffolded Question interactive", () => {
       cy.wait(200);
 
       getAndClearAllPhoneMessage((messages) => {
-        expect(messages.length).eql(3);
+        expect(messages.length).eql(2);
 
         expect(messages[0]).eql({
           action: "focus in",
           data: {
-            focus_target: "textarea",
+            target_element: 'textarea',
+            target_type: 'textarea',
+            target_id: '',
+            target_name: '',
+            target_value: '',
             scaffolded_question_level: 1,
             subinteractive_url: "/open-response",
             subinteractive_type: "open_response",
-            subinteractive_sub_type: undefined
+            subinteractive_sub_type: undefined,
+            subinteractive_id: 'int1'
           }
         });
 
         expect(messages[1]).eql({
           action: "focus out",
           data: {
+            target_element: 'textarea',
+            target_type: 'textarea',
+            target_id: '',
+            target_name: '',
+            target_value: 'Test subquestion answer',
             scaffolded_question_level: 1,
             subinteractive_url: '/open-response',
             subinteractive_type: 'open_response',
-            subinteractive_sub_type: undefined
-          }
-        });
-
-        expect(messages[2]).eql({
-          action: "answer saved",
-          data: {
-            answer_text: "Test subquestion answer",
-            scaffolded_question_level: 1,
-            subinteractive_url: "/open-response",
-            subinteractive_type: "open_response",
-            subinteractive_sub_type: undefined
+            subinteractive_sub_type: undefined,
+            subinteractive_id: 'int1',
+            answer_text: 'Test subquestion answer'
           }
         });
       });
--- a/cypress/integration/scaffolded-question.test.js
+++ b/cypress/integration/scaffolded-question.test.js
@@ -1,6 +1,6 @@
 import { phonePost, phoneListen, getAndClearLastPhoneMessage } from "../support";
 
-context("Test open response interactive", () => {
+context("Test Scaffolded Question interactive", () => {
   beforeEach(() => {
     cy.visit("/wrapper.html?iframe=/scaffolded-question");
   });
@@ -79,7 +79,8 @@ context("Test open response interactive", () => {
               answerType: "open_response_answer",
               answerText: "Test subquestion answer"
             }
-          }
+          },
+          answerText: "[Level: 1] Test subquestion answer"
         });
       });
     });
--- a/cypress/integration/scaffolded-question.test.js
+++ b/cypress/integration/scaffolded-question.test.js
@@ -1,4 +1,4 @@
-import { phonePost, phoneListen, getAndClearLastPhoneMessage } from "../support";
+import { phonePost, phoneListen, getAndClearLastPhoneMessage, getAndClearAllPhoneMessage } from "../support";
 
 context("Test Scaffolded Question interactive", () => {
   beforeEach(() => {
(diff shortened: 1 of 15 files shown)


pjanik:
My current implementation limits number of focus in/out events. E.g. when the user keeps clicking radio buttons in MC question, I'll log only one focus in, and focus out (+ answer saved) when the user completely leaves the question and focuses on the element outside the iframe (or non-input element in the iframe div). I think this is closer to what we had in LARA. Alternatively, we could log focus in/out (+ answer save) each time user changes a choice. 
Not sure what's better.

src/shared/components/base-question-app.tsx
```diff
@@ -54,6 +56,13 @@ export const BaseQuestionApp = <IAuthoredState extends IAuthoringMetadata & IBas
   useHint();
   useRequiredQuestion();
   useShutterbug({ container: "." + css.runtime });
+  const logging = useBasicLogging({ disabled: disableBasicLogging || !isRuntimeView });
+
+  const setInteractiveStateWithLogging = (updateFunc: UpdateFunc<IInteractiveState>) => {
+    setInteractiveState(updateFunc);
+    // Notify logging helper that answer has been updated by user. Only then it'll be logged on blur.
+    logging.onAnswerUpdate(updateFunc(interactiveState).answerText);
```
scytacki, line 64: I think this means the update function is called twice.
I don't have all of this code in my head, but I'd think that might be confusing to an interactive developer when calling setInteractiveState.

But I can't think of a simple way to fix that since we don't know when the updateFunc will be called by setInteractiveState.

src/shared/components/submit-button.tsx
```diff
@@ -20,6 +20,7 @@ export const SubmitButton: React.FC<IProps> = ({ isAnswered }) => {
 
   const handleSubmit = () => {
     setInteractiveState?.(prevState => ({...prevState, submitted: true }));
```
scytacki, line 22: So in this case, with an open response will there be an 'answer saved' event and a 'answer submitted' event since the focus would have left the the text box as the user clicks the submit button?
And in that case which one happens first?  
I'm wondering about the order because it seems possible the `submitted: true` appears to be unset in the second event if the order is a certain way.

src/shared/hooks/use-basic-logging.ts
```diff
@@ -0,0 +1,51 @@
...
+      }
+    };
+    const focusOut = (e: FocusEvent) => {
+      // event.relatedTarget === null means that user is most likely leaving to outer window, and not changing focus
+      // within the interactive iframe (eg clicking various buttons). We want to log only "main" blur event.
+      if (!e.relatedTarget) {
+        log("focus out");
+        if (answerToLog.current !== undefined) {
+          // LARA uses "answer saved" message for its basic questions too.
+          log("answer saved", { answer_text: answerToLog.current });
```
scytacki, line 31: For a multiple choice will the researcher be able to tell the id of the choice made by the student?

If the user closes the browser window is this answer saved event logged? 
I'd think that it might work sometimes but not others, I'd think closing the page might trigger a focus out, but the browser might kill off the code before it can send any messages.

cypress/integration/scaffolded-question.test.js
```diff
@@ -79,6 +79,71 @@ context("Test open response interactive", () => {
...
+        expect(messages.length).eql(3);
+
+        expect(messages[0]).eql({
+          action: "focus in",
+          data: {
+            focus_target: "textarea",
+            scaffolded_question_level: 1,
+            subinteractive_url: "/open-response",
+            subinteractive_type: "open_response",
+            subinteractive_sub_type: undefined
```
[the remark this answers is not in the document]
pjanik: I thought mostly about interactives like fill in the blank. Its type is "iframe_interactive", so it's not very descriptive (since we use the same metadata here). sub_type will be "fill_in_the_blak" (I don't think it's set yet), so it should make it a bit easier to filter events and provide more context. Does it make sense?
And right, I'll add subinteractive_id.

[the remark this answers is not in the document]
scytacki: Yeah, that makes sense. So it is fine to leave it.

src/shared/components/submit-button.tsx
pjanik: > I'm wondering about the order because it seems possible the submitted: true appears to be unset in the second event if the order is a certain way.

We don't log interactive state anyway, so submitted value isn't visible in the logs.
If you type in the box, move your mouse directly to Submit button, and click, the order of events is:
- "answer submitted", answer_text: "test"
- "focus out"
- "answer saved", answer_text: "test"

Generally saving on blur / focus out isn't perfect. I think I could check if there's submit=true in interactive state and don't trigger "answer saved" even then.

scytacki:
I just realized I didn't enable save state on the multiple choice interactive that I added that is why it wasn't saving. Sorry for the confusion. I'll remove it from the comment above.

src/shared/components/submit-button.tsx
scytacki: Nah, it is probably better to keep the answer saved logic simple so there is less chances for failure. 
Yeah the focus tracking isn't great, but I don't know how else to generically track user interaction.

src/shared/hooks/use-basic-logging.ts
pjanik: > For a multiple choice will the researcher be able to tell the id of the choice made by the student?

No, just the choice text + (correct) (if choice is marked as correct). Will that be a problem? Choice IDs can be long and not very readable (default 3 choices have simple, predefined IDs, but when you add a new one, I think UUID string is used). If we want to have a choice ID as a fallback, we could pass an interactive state too (but that would increase amount of data we send to log system).

> If the user closes the browser window is this answer saved event logged?

I tested that by replacing LOGGER_URI with my local server. It won't save it in practice. It's limitation of saving on blur, but that was a requirement. To make it more reliable, we would probably need to use the same approach that LARA is using to prevent data loss. So, I guess we'd need to make LARA log "answer saved" events instead of interactives. But then we would have more events in the stream and they won't be related to blur/focusout events. In this case, we could add answer_text to "focus out" event here, so researcher would have a choice which kind of events to use.

src/shared/components/base-question-app.tsx
pjanik: Yes, it'll be called twice.

But looking at your concerns from other comments, I think I'd vote for removing "answer saved" event and adding "answer_text" to "focus out". In this case, we might even skip this logic and simply always log answer_text on focus out.

Then, LARA could log "answer saved" when the answer is really saved.

pjanik:
> I do think we should be logging each time the student changes their choice in a multiple choice. Looking at the code I don't see why that doesn't happen, but I see you know about it based on your comment.

Done in: https://github.com/concord-consortium/question-interactives/pull/32/commits/148ac85e7fe390784b9d6c0ffa598543daa33ef8
I've also added much more properties to focus in/out so it should be easier to identify inputs.

I think I addressed other issues in new commits too. I removed "answer saved" event as it creates a false impression that the answer really has been saved at this moment. Instead, I just added "answer_text" to "focus out", as that's what in fact we're logging. It feels to me that "answer saved" should be added to LARA if it's necessary.

pjanik:
I've added "scrolled into view" and "scrolled out of view" events in the last commit. It's separate story, but I used the same hook, so it made sense to keep it together. IntersectionObserver seems to have good enough support not to use any polyfill: https://caniuse.com/#feat=intersectionobserver
But if I'm missing something (Safari < 12?), let me know and I'll check some polyfills.

scytacki:
We'll get by without the choice id. The main consumer of this kind of log data will be other programs making reports. However before those programs are written, the log data is viewed by researchers who don't necessarily think about what it is going to take to write a program.  So really we need both pieces of data.

kswenson:
There does appear to be a heavily-used polyfill: https://www.npmjs.com/package/intersection-observer.

PR #36, opened 2020-07-02
Replace TSLint with ESLint and fix multiple issues (mostly hooks-related)

pjanik:
This shows why we should use ESLint and this new config. 😉 
I'd really recommend migrating other TS and hook-based projects (probably Vortex would be the best candidate, as it's using hooks heavily and it's relatively complex). I think that the most common, false assumption is that `useEffect(..., []) == componentDidMount`. 

The most important changes are in scaffolded-question/iframe-runtime.tsx and in the video-player/runtime.tsx. Hopefully, comments should make them more clear.

@scytacki, I've made some changes to useAutoHeight hook that might affect your sizing fix work (and probably conflict with currently open PR).

I did some manual testing, but I think a solid QA might need to be necessary after this PR is merged.


src/scaffolded-question/components/iframe-runtime.tsx
```diff
@@ -27,67 +27,68 @@ export const IframeRuntime: React.FC<IProps> =
   const [ hint, setHint ] = useState("");
   const iframeRef = useRef<HTMLIFrameElement>(null);
   const phoneRef = useRef<IframePhone>();
+  // Why is interativeState and setInteractiveState kept in refs? So it's not necessary to declare these variables as
+  // useEffect's dependencies. Theoretically this useEffect callback is perfectly fine either way, but since
```
kswenson, line 31: A cleaner solution -- if only it were supported -- would be an eslint comment that would make it possible to specify that a particular value is intentionally left out of the dependencies list. In the absence of that, this seems like a reasonable alternative.

src/setupTests.js
```diff
@@ -1,4 +1,6 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
kswenson, line 1:
```suggestion
// eslint-disable @typescript-eslint/no-var-requires
```
I think you could reasonably make it a single file-wide configuration.

src/video-player/components/runtime.tsx
```diff
@@ -37,36 +37,25 @@ export const getAspectRatio = (aspectRatio: string) => {
...
-  };
-  const getViewPercentage = () => {
-    if (playerRef.current) {
-      return playerRef.current.currentTime / playerRef.current.duration;
-    }
-    else return 0;
-  };
 
-  const loadPlayer = () => {
+  // Keep viewedTimestamp in ref, so the useEffect below doesn"t need to list it in its dependency array, and doesn"t
```
kswenson, line 46:
```suggestion
  // Keep viewedTimestamp in ref, so the useEffect below doesn"t need to list it in its dependency array, and doesn't
```

src/video-player/components/runtime.tsx
```diff
@@ -37,36 +37,25 @@ export const getAspectRatio = (aspectRatio: string) => {
...
+  // get triggered each time viewedTimestamp is updated. Its callback treats this value as an initial value.
+  // We don"t want to reload video player each time new timestamp is saved.
```
kswenson, line 48:
```suggestion
  // We don't want to reload video player each time new timestamp is saved.
```

src/scaffolded-question/components/iframe-runtime.tsx
```diff
@@ -27,67 +27,68 @@ export const IframeRuntime: React.FC<IProps> =
...
+  // it reloads the iframe each time it's called, it's not a great experience for user when that happens while he is
+  // interacting with the iframe (e.g. typing in textarea). And interactiveState is being updated very often,
+  // as well as setInteractiveState that is generated during each render of the parent component.
+  const interactiveStateRef = useRef<any>(undefined);
+  const setInteractiveStateRef = useRef<((state: any) => void) | undefined>(undefined);
```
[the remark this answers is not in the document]
pjanik: Thanks, I'll do it, as then I could skip `| undefined` typing.

src/scaffolded-question/components/iframe-runtime.tsx
```diff
@@ -27,67 +27,68 @@ export const IframeRuntime: React.FC<IProps> =
...
-        subinteractive_type: authoredState.questionType,
-        subinteractive_sub_type: authoredState.questionSubType,
-        subinteractive_id: id
+  useEffect(() => {
+    const initInteractive = () => {
+      const phone = phoneRef.current;
+      if (!phone) {
+        return;
+      }
+      phone.addListener("interactiveState", (newInteractiveState: any) => {
```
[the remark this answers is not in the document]
pjanik: I don't think so. This component is a generic container of any subinteractive in the scaffolded question. It doesn't make any assumption about `interactiveState` (and `authoredState`), it just passes it around. So, it's really `any` type. ;)

pjanik:
Thanks. I've addressed @kswenson comments.

scytacki:
Since Kirk already approved. I'm going to merge this in.

PR #37, opened 2020-07-06
Change rendering of credit links on images in lightbox

dstrawberrygirl:
Tweak the rendering so that images that are in the lightbox have all text on one line


src/image/components/runtime.tsx
```diff
@@ -50,21 +50,26 @@ export const Runtime: React.FC<IProps> = ({ authoredState, report }) => {
...
-      <div className={css.creditLink}>
-        <a href={creditLink} target="_blank" rel="noopener">
-          {creditLinkDisplayText || creditLink}
-        </a>
-      </div>
-    )
+    const link = <a href={creditLink} target="_blank" rel="noopener">
+                  {creditLinkDisplayText || creditLink}
+                </a>
+    if (displayBlock === true) {
```
kswenson, line 64:
```suggestion
    if (displayBlock) {
```

PR #38, opened 2020-07-07
Demo: Image sizing tweaks and documentation

scytacki:
This PR is documentation of a path we shouldn't take, rather than something to be merged.

The point of this work on aspectRatio is so LARA can use its sizing calculations to make sure the interactive doesn't get bigger than the viewable area of the screen.  When an interactive sets its aspectRatio LARA does these calculations, when the interactive sets its height LARA doesn't do these calculations. 

It is best leaving image and video interactives working as they currently are which means they set their height and LARA can't limit their size to make sure they fit.  The current built in (non iframe) image interactive doesn't do any size limiting either, so this won't be a regression if we leave it as is.

The comments in the code document the sizing tweaks necessary for the aspectRatio to sort of work.

This also conditionally disables the autoHeight based on the authored settings of the
image interactive. This way the aspectRatio fudging is only used when it is really needed.
Conditionally disabling the autoHeight required required passing the container reference
into the Runtime component. This might cause a performance problem. It also broke the tests.

The better solution for all of this is for LARA to notify the interactive when the max viewable
height has changed. Then the interactive can decide if it wants to limit the height it
sends to LARA based on this max viewable height. The interactive won't have a way
to adjust its width, but currently this kind of width adjustment is not really visible to the user.  The only time the actual width of the interactive would only be visible is when the interactive is wrapped with a teacher edition plugin. That plugin adds a border around the interactive so that border would be bigger than necessary if the interactive letter boxed its contents. That kind of extra space does not seem like a problem to me, so it I don't think the width adjustment is necessary.

[#173174724]

### DIFF
--- a/src/image/components/runtime.tsx
+++ b/src/image/components/runtime.tsx
@@ -1,4 +1,4 @@
-import React, { useRef } from "react";
+import React, { useRef, RefObject } from "react";
 import { IAuthoredState } from "./app";
 import { setSupportedFeatures, showModal } from "@concord-consortium/lara-interactive-api";
 import ReactDOMServer from "react-dom/server";
@@ -6,9 +6,11 @@ import { v4 as uuidv4 } from "uuid";
 import { log } from "@concord-consortium/lara-interactive-api";
 import css from "./runtime.scss";
 import ZoomIcon from "../../shared/icons/zoom.svg";
+import { useAutoHeight } from "../../shared/hooks/use-auto-height";
 
 interface IProps {
   authoredState: IAuthoredState;
+  baseContainer: RefObject<HTMLDivElement>;
   report?: boolean;
 }
 
@@ -17,9 +19,22 @@ interface IImageSize {
   height: number;
 }
 
-export const Runtime: React.FC<IProps> = ({ authoredState, report }) => {
+export const Runtime: React.FC<IProps> = ({ authoredState, baseContainer, report }) => {
 
   const imageSize = useRef<IImageSize>();
+
+  const creditTextContainer = useRef<HTMLDivElement>(null);
+
+  // We've disabled the autoHeight hook provided by the base app
+  // When the scaling is fitWidth then the image change its size based on the
+  // width of the container, so we should use LARA aspect raio setting. This
+  // way the image's width and height will be limited so the whole thing shows
+  // within the viewport.
+  // When the scaling is originalDimensions then it is best to use autoHeight
+  // because the height of the interactive is fixed based on the image height
+  // the image does not resize in this case so this is perfect for using LARA's
+  // height option
+  useAutoHeight({ container: baseContainer, disabled: authoredState.scaling === "fitWidth" });
 
   const getImageLayout = () => {
     switch (authoredState.scaling) {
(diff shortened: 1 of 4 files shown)


scytacki:
This PR was based off of this one: https://github.com/concord-consortium/question-interactives/pull/27

scytacki:
@dstrawberrygirl  and @kswenson you should probably take a look at this.

@pjanik you might also be interested since it shows a hacky way to work with the useAutoHeight hook.

But again this is more documentation of what we probably shouldn't do.

src/image/components/runtime.tsx
```diff
@@ -33,17 +48,88 @@ export const Runtime: React.FC<IProps> = ({ authoredState, report }) => {
...
+
+      if (authoredState.scaling === "fitWidth") {
+        if (!baseContainer.current) {
+          // FIMXE:
+          // If the image loads before the baseContainer is initialized this will
+          // default to LARA default aspect ratio of 1.3
+          // That will result in a clipped image in most cases.
+          // A better solution would be to delay the execution of this method
+          // until both the image is loaded and the baseContainer is available
+          console.error("image interactive baseContainer not available. aspectRatio will be wrong");
```
pjanik, line 60: Have you ever had this problem / seen this error message?
I guess this check won't hurt and might help to debug some issues in the future. But I'll be surprised to see it.

src/image/components/runtime.tsx
```diff
@@ -33,17 +48,88 @@ export const Runtime: React.FC<IProps> = ({ authoredState, report }) => {
...
+        // loop:
+        //   If the interactive is in the interactive box and the user resizes
+        //   the window then the width of the interactive will change. This would
+        //   trigger a re-calcuation of the aspectRatio in the interactive.
+        //   When the interactive sends this new aspectRaio to LARA,
+        //   LARA will re-calcuate the width and height of the interactive again.
+        //   LARA's calculation of the width will change if the
+        //   users browser is short so maintainig the aspect ratio means limiting
+        //   the width.  This width change will trigger the interactive to send a
+        //   new aspect ratio. This will go on and on.
```
pjanik, line 83: Have you tried this approach in practice?
I agree that it's worth it, I'm just curious. I could imagine the loop should automatically stop itself. And/or a reasonable stop condition could be not to send a new aspectRatio to LARA if it's very close to the current one.

src/image/components/runtime.tsx
```diff
@@ -17,10 +19,23 @@ interface IImageSize {
...
+  // We've disabled the autoHeight hook provided by the base app
+  // When the scaling is fitWidth then the image change its size based on the
+  // width of the container, so we should use LARA aspect raio setting. This
+  // way the image's width and height will be limited so the whole thing shows
+  // within the viewport.
+  // When the scaling is originalDimensions then it is best to use autoHeight
+  // because the height of the interactive is fixed based on the image height
+  // the image does not resize in this case so this is perfect for using LARA's
+  // height option
+  useAutoHeight({ container: baseContainer, disabled: authoredState.scaling === "fitWidth" });
```
pjanik, line 37: Wouldn't `useAutoHeight` work with the main container (div) defined in this component? 
It could simplify things a bit, as we wouldn't have to pass the ref. But it's not really important.

Another option would be to use `document.body` in `useAutoHeight` and get rid of `container` option (or make it optional). I guess that's what we usually want.

src/image/components/runtime.tsx
scytacki: Yeah I didn't see, but it seemed theoretically possible, so I added the error just incase.

src/image/components/runtime.tsx
scytacki: I haven't tried it with this interactive. 

But we did try something like this before we had the setHeight api and the interactive continued to jitter around. Partly because the height calculation resets it to 100% width before each calculation. So it isn't the case that we are getting closer and closer to some final state. Instead from the interactives point of view it is jumping to a large width, then jumping back to a close to the fine width, then jumping to a large width, ...

src/image/components/runtime.tsx
scytacki: I tried using the div defined in the component, but that didn't include the margins that are applied by the base app. So then it wasn't request enough height.

Yeah having useAutoHeight default to the document.body, would make this better. That'd also fix the testing issue since the testing framework wasn't happy initializing a ref inside of the test.

scytacki:
I'm going to close this, because I don't think we should merge this.

If we do come back to it, and need to customize autoHeight, then making the change that Piort suggested of having useAutoHeight default to `document.body`, would simplify this approach.

The best approach is documented above. The summary is:
- LARA should send the max viewable height to the interactive and re-send it when ever it changes. This can happen when the user resizes the browser, or it could happen if there is some part of the page that takes of vertical space and this part of the page is changing size.
- Now the interactive can take this max viewable height into account and adjust the height it requests from LARA based on it.

This allow the interactive limit its own height if it wants to. It also simplifies the calculation at the interactive level and cuts out any loops between the interactive and LARA.

PR #39, opened 2020-07-08
Drawing tool

sfentress:
This adds the drawing tool, with authoring for background images and stamps collections.

The drawing tool can be seen here: https://lara-master.concordqa.org/activities/20000/pages/304600/edit

@pjanik I was a little confused about the types for the authoring state. I made it `IAuthoringOpenResponseMetadata` which made `BaseQuestionApp` happy. If I tried to make it the more generic type `IAuthoringMetadataBase` it complained. This seems fine, except that the tests then require the authored data to contain `questionType: "open_response"`, which is a little weird.

I set the actual questionType to `drawing_tool`, also unsure if that's correct.

### DIFF
--- a/cypress/integration/drawing-tool.test.js
+++ b/cypress/integration/drawing-tool.test.js
@@ -1,0 +1,51 @@
+import { phonePost, phoneListen, getAndClearLastPhoneMessage } from "../support";
+
+const authoredStateWithoutStamps = {
+  version: 1,
+  questionType: "open_response",
+  imageFit: "shrinkBackgroundToCanvas",
+  imagePosition: "center",
+  stampCollections: []
+};
+
+const authoredStateWithStamps = {
+  version: 1,
+  questionType: "open_response",
+  imageFit: "shrinkBackgroundToCanvas",
+  imagePosition: "center",
+  stampCollections: [{
+    collection: "custom",
+    name: "My stamps",
+    stamps: [
+      "https://interactions-resources.concord.org/stamps/simple-atom.svg"
+    ]
+  }]
+};
+
+context("Test Image interactive", () => {
+  beforeEach(() => {
+    // cy.viewport(800, 600);
+    cy.visit("/wrapper.html?iframe=/drawing-tool");
+  });
+
+  context("Runtime view", () => {
+    it("renders drawing tool without stamps", () => {
+      phonePost("initInteractive", {
+        mode: "runtime",
+        authoredState: authoredStateWithoutStamps
+      });
+
+      cy.getIframeBody().find(".dt-container").should('exist');
+      cy.getIframeBody().find('[title="Stamp tool (click and hold to show available categories)"]').should('not.exist');
+    });
+
+    it("renders drawing tool with stamps", () => {
+      phonePost("initInteractive", {
+        mode: "runtime",
+        authoredState: authoredStateWithStamps
+      });
+
+      cy.getIframeBody().find('[title="Stamp tool (click and hold to show available categories)"]').should('exist');
+    });
+  });
+});
--- a/cypress/integration/drawing-tool.test.js
+++ b/cypress/integration/drawing-tool.test.js
@@ -1,4 +1,4 @@
-import { phonePost, phoneListen, getAndClearLastPhoneMessage } from "../support";
+import { phonePost } from "../support";
 
 const authoredStateWithoutStamps = {
   version: 1,
--- a/cypress/integration/drawing-tool.test.js
+++ b/cypress/integration/drawing-tool.test.js
@@ -2,7 +2,7 @@ import { phonePost, phoneListen, getAndClearLastPhoneMessage } from "../support"
 
 const authoredStateWithoutStamps = {
   version: 1,
-  questionType: "open_response",
+  questionType: "iframe_interactve",
   imageFit: "shrinkBackgroundToCanvas",
   imagePosition: "center",
   stampCollections: []
@@ -10,7 +10,7 @@ const authoredStateWithoutStamps = {
 
 const authoredStateWithStamps = {
   version: 1,
-  questionType: "open_response",
+  questionType: "iframe_interactve",
   imageFit: "shrinkBackgroundToCanvas",
   imagePosition: "center",
   stampCollections: [{
--- a/src/drawing-tool/components/app.tsx
+++ b/src/drawing-tool/components/app.tsx
@@ -125,6 +125,10 @@ const baseAuthoringProps = {
         title: "Prompt",
         type: "string"
       },
+      required: {
+        title: "Required (Show submit and lock button)",
+        type: "boolean"
+      },
       hint: {
         title: "Hint",
         type: "string"
--- a/src/drawing-tool/components/app.tsx
+++ b/src/drawing-tool/components/app.tsx
@@ -22,7 +22,9 @@ export interface IAuthoredState extends IAuthoringInteractiveMetadata {
   stampCollections: StampCollection[];
 }
 
-export interface IInteractiveState extends IRuntimeInteractiveMetadata {}
+export interface IInteractiveState extends IRuntimeInteractiveMetadata {
+  drawingState: string;
+}
 
 const baseAuthoringProps = {
   schema: {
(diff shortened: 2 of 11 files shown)


src/drawing-tool/components/runtime.tsx
```diff
@@ -0,0 +1,109 @@
...
+  width: number;
+  height: number;
+  stamps?: StampCollections;
+}
+
+export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, setInteractiveState, report }) => {
+
+  const readOnly = !!(report || (authoredState.required && interactiveState?.submitted));
+
+  // equivalent to componentDidLoad
```
pjanik, line 33: This statement isn't 100% true. Sounds like nitpicking, but I've already seen problems related to this assumption. It's pretty common when we transition from Class-based components. If you merge in `master` branch, you should see quite a few errors/warning from ESLint (since TSLint has been replaced with ESLint with rules-of-hooks plugins in master). It could be fixed by adding dependencies to useEffect's second argument.

src/drawing-tool/components/runtime.tsx
```diff
@@ -0,0 +1,109 @@
...
+    if (interactiveState) {
+      drawingTool.load(interactiveState.answerText, null, true);
+    }
+
+    drawingTool.on('drawing:changed', () => {
+      if (readOnly) return;
+      const userState = drawingTool.save();
+      setInteractiveState?.(prevState => ({
+        ...prevState,
+        answerText: userState,
```
pjanik, line 94: `answerText` shouldn't be used for keeping actual state. `answerText` is actually meant to be meaningful text. It's used by various reports and logs to display an answer summary, without forcing teachers to load the whole iframe. Note that it's optional and you don't have to use it if it doesn't make sense for drawing tool.

src/drawing-tool/components/app.tsx
```diff
@@ -0,0 +1,216 @@
...
+      },
+    },
+    properties: {
+      version: {
+        type: "number",
+        default: 1
+      },
+      questionType: {
+        type: "string",
+        default: "drawing_tool"
```
pjanik, line 122: I don't think it's a valid type. I guess you should use "iframe_interactive" as its the default one. This is important for reporting. If you used "open_response" or "multiple_choice", LARA and Teacher Report would expect certain fields and they would report this question as open_response/multiple_choice, etc.

cypress/integration/drawing-tool.test.js
```diff
@@ -0,0 +1,51 @@
...
+  version: 1,
+  questionType: "open_response",
+  imageFit: "shrinkBackgroundToCanvas",
+  imagePosition: "center",
+  stampCollections: []
+};
+
+const authoredStateWithStamps = {
+  version: 1,
+  questionType: "open_response",
```
pjanik, line 13: I don't think "open_response" should be here. I think you want to use `"iframe_interactive"` (or `"iframe_interactive" as const` to make TypeScript happy).

cypress/integration/drawing-tool.test.js
```diff
@@ -0,0 +1,51 @@
+import { phonePost, phoneListen, getAndClearLastPhoneMessage } from "../support";
+
+const authoredStateWithoutStamps = {
+  version: 1,
+  questionType: "open_response",
```
pjanik, line 5: I don't think "open_response" should be here. I think you want to use `"iframe_interactive"` (or `"iframe_interactive" as const` to make TypeScript happy).

src/drawing-tool/components/runtime.tsx
```diff
@@ -0,0 +1,109 @@
...
+        stamps = predefinedStampCollections[collection.collection];
+      }
+      if (stamps && stamps.length) {
+        stampCollections[name] = stamps;
+      }
+    });
+
+    const drawingToolOpts: DrawingToolOpts = {
+      width: windowWidth - kToolbarWidth - 10,
+      height: kToolbarHeight
```
pjanik, line 60: Interactive can be resized by LARA (eg. in the responsive layout). I guess you might need to use https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver to monitor changes to iframe dimensions (it's already used in this repo). Not sure if DrawingTool will like dynamic resizing. I guess it might be a new feature there.

src/drawing-tool/components/runtime.tsx
sfentress: 👍 . This was from thinking that I had to use `IAuthoringOpenResponseMetadata`

src/drawing-tool/components/app.tsx
sfentress: 👍

sfentress:
Thanks for the thorough review. Types and question types are updated -- it wasn't clear that `iframe_interactive` was supposed to be a generic type even for standardized questions like the draw tool, I had figured that that was for things like CBio.

I'm happy to go with the updated version of using refs within `useEffect` to not trigger changes.

src/drawing-tool/components/runtime.tsx
```diff
@@ -0,0 +1,118 @@
...
+  width: number;
+  height: number;
+  stamps?: StampCollections;
+}
+
+export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, setInteractiveState, report }) => {
+
+  const readOnly = !!(report || (authoredState.required && interactiveState?.submitted));
+
+  // need a wrapper as `useRef` expects (state) => void
```
pjanik, line 33: I guess you could just specify setInteractiveStateRef type as: `useRef<(updateFunc: (prevState: IInteractiveState | null) => IInteractiveState) => void>` (same as above). But it doesn't seem to matter.

pjanik:
D'oh, I've just noticed I'm reviewing merged PR.

sfentress:
Hi @pjanik 

> I think you'll have problems with sizing, especially if you try this interactive in responsive layout

Yes, but the team didn't want to address this in the first round. It seems like it would be better to address this at the drawing tool level, to allow a mode that is responsive. As it is now, re-initializing the drawing tool when the window changes width is messy. It's also not clear what should happen to the user data. Currently the width is saved to the user data. Should a researcher see a drawing tool at whatever width their own window is at, or how the student left it? etc.

> test coverage hasn't change

Test coverage includes basic proof of loading for authored states. Making a cypress test for actually drawing on the canvas seemed pretty tricky, as did one of loading student data, since there are no dom elements to check for. But if you have ideas I'm happy to hear.

> Also, you specified isAnswered function to check answerText property.

Ack, that's a bug. Will fix. Good catch.

> And the last thing - when readOnly gets updated to true (after you submit a question), it seems we'll reinitialize a DrawingTool. 

This seemed to work fine.

> D'oh, I've just noticed I'm reviewing merged PR.

No problem. I had merged it in as I weren't sure if you were going to take a second look.

sfentress:
> Ack, that's a bug. Will fix. Good catch.

A fix for this is in #44, I won't merge yet in case you have further thoughts for the above.

pjanik:
> Yes, but the team didn't want to address this in the first round. It seems like it would be better to address this at the drawing tool level, to allow a mode that is responsive. As it is now, re-initializing the drawing tool when the window changes width is messy. It's also not clear what should happen to the user data. Currently the width is saved to the user data. Should a researcher see a drawing tool at whatever width their own window is at, or how the student left it? etc.

Yeap, this doesn't seem to be straightforward to fix, it makes sense to leave it for later. Just wanted to warn. ;) 

> Making a cypress test for actually drawing on the canvas seemed pretty tricky, as did one of loading student data, since there are no dom elements to check for. But if you have ideas I'm happy to hear.

Right, I wouldn't expect drawing in Cypress (it could be fun actually if there's enough time ;)). If you look at the initial questions like open-response, fill-in-the-blank, etc., they have pretty excessive tests and Runtime component has 100% coverage. Some are almost unnecessary - like Cypress tests authoring interface that has been unified in the meantime and provided by the same component everywhere. But actually they already helped me when I was reworking the repo. And ofc I imagine that open-response textarea is easier to test than drawing tool. ;) But if the time is limited and things get more complex, I think the most important tests that every question should have is:
- loading authored state (already there)
- loading student state (interactive state)
- making sure that student state (interactive state) gets saved

It doesn't have to be Cypress, it could be Jest tests of runtime.tsx and DrawingTool could be mocked using Jest. Not saying this has to be part of this PR, I imagine there is some schedule and/or deadlines. But in general, I'd try to sneak that into future stories/estimations. ;)

PR #40, opened 2020-07-09
Fix cypress tests

kswenson:
Cypress tests have been somewhat inexplicably failing on master and on several PR branches for reasons that don't seem obviously related to any recent code changes. After working with @eireland for a while this afternoon/evening, the following changes seem to be sufficient, although it's not entirely clear which are necessary:
- update to cypress 4.10
- add waits before page visits
- video interactive waits for video to finish before querying received messages
- video interactive rearranges the order of some assertions

### DIFF
--- a/cypress/integration/multiple-choice.test.js
+++ b/cypress/integration/multiple-choice.test.js
@@ -1,7 +1,10 @@
 import { phonePost, phoneListen, getAndClearLastPhoneMessage } from "../support";
 
 context("Test multiple-choice interactive", () => {
+  let i = 0;
+
   beforeEach(() => {
+    if (i++ > 0) cy.wait(3000);
     cy.visit("/wrapper.html?iframe=/multiple-choice");
   });
 
@@ -392,7 +395,7 @@ context("Test multiple-choice interactive", () => {
       cy.getIframeBody().find("#root_multipleAnswers").click();
       getAndClearLastPhoneMessage(state => {
         expect(state.multipleAnswers).eql(true);
-      });
+      }, 100);
 
       cy.getIframeBody().find(".btn-add").click();
       cy.getIframeBody().find("#root_choices_0_content").clear();
@@ -401,7 +404,7 @@ context("Test multiple-choice interactive", () => {
       getAndClearLastPhoneMessage(state => {
         expect(state.choices[0].content).eql("Choice A");
         expect(state.choices[0].correct).eql(true);
-      });
+      }, 100);
 
       cy.getIframeBody().find(".btn-add").click();
       cy.getIframeBody().find("#root_choices_1_content").clear();
@@ -409,7 +412,7 @@ context("Test multiple-choice interactive", () => {
       getAndClearLastPhoneMessage(state => {
         expect(state.choices[1].content).eql("Choice B");
         expect(state.choices[1].correct).eql(false);
-      });
+      }, 100);
     });
   });
 
--- a/cypress/integration/video-player.test.js
+++ b/cypress/integration/video-player.test.js
@@ -14,7 +14,10 @@ const authoredStateSample = {
 
 
 context("Test Video Player interactive", () => {
+  let i = 0;
+
   beforeEach(() => {
+    if (i++ > 0) cy.wait(3000);
     cy.visit("/wrapper.html?iframe=/video-player");
   });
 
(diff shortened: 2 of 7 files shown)


cypress/integration/video-player.test.js
```diff
@@ -63,14 +66,14 @@ context("Test Video Player interactive", () => {
...
       getAndClearLastPhoneMessage((state) => {
         expect(state.lastViewedTimestamp).greaterThan(1.0);
-
-        cy.getIframeBody().find(".vjs-text-track-cue").should("include.text", "This is a drake");
-        cy.getIframeBody().find(".vjs-current-time-display").should("include.text", "0:01");
-        cy.getIframeBody().find(".vjs-poster.vjs-hidden").should("exist");
-
-      });
+        // must wait for video to finish apparently
+      }, 15000);
```
pjanik, line 76: If I remember video implementation correctly, it should update progress every second. So, theoretically, it'd be enough to wait 1-2 seconds max. I guess you tried smaller values before? When I was testing that in practice in LARA, I wasn't sure if it actually works as designed.

cypress/integration/multiple-choice.test.js
```diff
@@ -1,7 +1,10 @@
 import { phonePost, phoneListen, getAndClearLastPhoneMessage } from "../support";
 
 context("Test multiple-choice interactive", () => {
+  let i = 0;
+
   beforeEach(() => {
+    if (i++ > 0) cy.wait(3000);
```
pjanik, line 7: These wait times are pretty long. Is that the smallest value that helps?

kswenson:
>It seems to fix the problem, so it's nice. Things seemed to work in a consistent way up to some point, so it'd be great to investigate changes that made tests failing for the first time.

I completely agree and I spent a while trying to figure it out. I could not identify any changes to our code that could reasonably have such wide-ranging effects nor could I find any recent dependency changes that seemed like plausible explanations. Some change to cypress itself seemed like the most likely explanation, so I tried every version from 3.8 to 4.10 without finding one that behaved differently. I wish we had a better explanation but having spent more time than I could reasonably afford down that rabbit hole I chose to move on to other things. I should have included more of this information in the original PR, so thanks for asking.

My best guess is that there are multiple things going on that is confusing the debugging process. There are at least two or three separate things at work here, and the fact that they all seem to have occurred at once may just be an odd coincidence, although that seems difficult to believe as well.

1. Something Electron-related -- the need for the 3-4 second wait around each visit is Electron-only. It shows up locally and on Travis and whether headless or not. Neither Chrome nor Firefox have any problem. Unfortunately, Electron is the only option on Travis with the trusty distribution. (It is possible to install Chrome on xenial via Travis, but we can't use xenial because of the `s3_website` dependency. I tried to figure out a way to install chrome "manually" on trusty without success.) I believe that the version of Electron is determined by the version of cypress, which is why I tried every version of cypress from 3.8 on up, but it's possible that at the time of that experiment I was being led astray by one of the other issues. One could in theory control the waits via an environment variable and run the cypress tests locally using Chrome without the waits and then on Travis with Electron and the waits, but that would just mean that a local cypress run would give less indication of whether a Travis build would succeed, so I'm not sure it's worth it.

2. The requirement for the 15-second video wait occurs in all browsers, so is not Electron-related. You asked about the 1-second callbacks that should occur. Those callbacks are occurring, but for some reason that I could not get to the bottom of, while the video is playing those callbacks go to the interactive library instead of the mock. This suggests that there's something flaky about the process of mocking postMessage listeners.

3. The additional smaller delays around calls to `getAndClearLastPhoneMessage()` were to address more intermittent failures that didn't seem obviously related to the first two problems.

The upshot is that nothing I encountered during my investigations pointed at anything wrong with our production code. Instead, everything pointed to something in our testing infrastructure, e.g. cypress, Electron, mocks, etc. In the end throwing in the waits and moving on seemed like the best option.

>Re cy.wait calls - are these minimal values that actually helped? As 3-4s before each test seem like quite a lot.

I did not conduct an extensive experiment, but my first tests were with one and two seconds, which failed, with three being the first value that allowed for a successful test run. Three worked fairly consistently but would occasionally fail on certain tests, in which case I bumped it up to four. It's possible that with additional effort they could be fine-tuned to test content, but I'm not sure it'd be worth it.

cypress/integration/multiple-choice.test.js
kswenson: As I mentioned elsewhere:

>I did not conduct an extensive experiment, but my first tests were with one and two seconds, which failed, with three being the first value that allowed for a successful test run. Three worked fairly consistently but would occasionally fail on certain tests, in which case I bumped it up to four. It's possible that with additional effort they could be fine-tuned to test content, but I'm not sure it'd be worth it.

pjanik:
Okay, thanks for the explanations. Right, I agree it makes sense to merge it and move on. Maybe we could just add some story to PT to optimize that one day (if we're lucky, new Electron/Cypress version might fix many of these problems).

PR #41, opened 2020-07-14
Call onChange from handleChange rather than handleBlur

kswenson:
- fixes bug with clicking save button without blurring the editor
- separately, uses `useCallback` to eliminate performance warning about unstable props to `<SlateEditor>`

Note: had to restart the cypress tests a couple of times to get past flaky test failures (multiple-choice once and scaffolded-question twice).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16466,6 +16466,11 @@
       "integrity": "sha512-cwESVXlO3url9YWlFW/TA9cshCEhtu7IKJ/p5soJ/gGpj7vbvFrAY/eIioQ6Dw23KjZhYgiIo8HOs1nQ2vr/oQ==",
       "dev": true
     },
+    "use-debounce": {
+      "version": "3.4.3",
+      "resolved": "https://registry.npmjs.org/use-debounce/-/use-debounce-3.4.3.tgz",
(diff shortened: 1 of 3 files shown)


pjanik:
Looks good, although not sure if I understand it fully. Is debouncing fixing some issues? Note that lara-interactive-api library also does it. Each time interactive state is updated, it'll wait 2 seconds before sending the state to LARA. This might sum up to 4 seconds now. Also, lara-interactive-api has special code that warns user that some data might be lost if user tries to change/reload page within these 2 seconds. Should we be worried about that here?

kswenson:
>Note that lara-interactive-api library also does it.

Ah, I didn't realize that -- or I had forgotten it. In that case I presumably don't need to worry about debouncing here. Thx for pointing it out. The change that fixes the bug was moving the `onChange` callback from the `onBlur` handler to the `onChange` handler. I'll update the PR accordingly.

kswenson:
@pjanik I removed the call to debounce. Ready for re-review.

kswenson:
Right, the larger point is that debouncing is not a rich-text-widget problem. If debouncing is warranted, it makes sense for it to happen at the lara-interactive-api level rather than having each individual widget debounce itself.

PR #42, opened 2020-07-17
Fix rich text dialog colors

kswenson:
- update to latest `@concord-consortium/slate-editor` library
- specify `themeColor` for rich text dialogs
- eliminate use of `tab()` for rich text cypress tests
  - it's no longer necessary since we no longer rely on blur events


pjanik:
Looks good to me. Not sure if I get the style changes though. Is `.modal-dialog` used for example by the color picker? And it clashes with our bootstrap styles that also specify `.modal-dialog`? 

If so, and if `.modal-dialog` is a descendant of `custom-rich-text-widget`, then we could probably do it a bit safer:

1. Keep `custom-rich-text-widget` as a regular CSS module, and use it in JSX like `css.customRichTextWidget` (as it seems to be defined in `rich-text-widget.tsx`, so it should work that way).

2. Don't make this CSS file global, and edit content a bit:
```
.customRichTextWidget {
  :global(.modal-dialog) {
     ...
  }
}
```
(`:global`  lets us define just one class name as global)

Ofc this only works if `.modal-dialog` is descendant of customRichTextWidget.

Otherwise, I don't know how to avoid conflicts with bootstrap. ;)

kswenson:
Thx for the tip. I'll look into it.

kswenson:
@pjanik Thx for pointing me towards `:global`. The fix ended up being a bit more elaborate than you suggested, but I think it's an improvement. I'm going to go ahead and merge this in the interest of pushing us closer to a testable release but feel free to review the changes in case there's anything that could still be improved.

PR #43, opened 2020-07-17
Log different events for the video player

dstrawberrygirl:
We were already logging the play / stop of the video player, this PR addresses the events relating to captions / subtitles, and handling the user scrubbing through a video as a separate logged event.

There are no easy hooks to capture the toggle of a caption so I'm relying on the regular state save interval to check on the current state and report if it has been changed. This keeps us from having to dig through the control bar hierarchy to find the actual toggle.

### DIFF
--- a/src/video-player/components/runtime.tsx
+++ b/src/video-player/components/runtime.tsx
@@ -83,8 +83,9 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
           // Store a ref to the captions track s we can check visibility later. There is no easy toggle event for captions show/hide
           captionsTrackRef.current = player.textTracks()[0];
           if (captionsTrackRef.current) {
-            captionsTrackRef.current.addEventListener("cuechange", (cue) => {
-              log("cue change", { videoUrl: authoredState.videoUrl });
+            captionsTrackRef.current.addEventListener("cuechange", () => {
+              const currentCue = captionsTrackRef.current?.activeCues?.length && captionsTrackRef.current?.activeCues.length > 0 && captionsTrackRef.current?.activeCues[0];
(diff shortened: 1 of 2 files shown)


src/video-player/components/runtime.tsx
```diff
@@ -116,32 +128,45 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
...
+      const percentageViewed = getViewPercentage();
+      saveStateInterval.current += 1;
+      setInteractiveState?.(prevState => ({
+        ...prevState,
+        answerType: "interactive_state",
+        answerText: `Percentage viewed: ${viewedProgress}`,
+        percentageViewed,
+        lastViewedTimestamp: getViewTime()
+      }));
+      if (logMessage.length > 0) {
```
sfentress, line 162: This reads slightly confusing to me, as I normally associate `.length` with array-like objects. Any reason not to have just `if (logMessage)` which would come to the same thing? And, indeed, the logMessage argument could be optional without the `""` default, (`logMessage: string?`), which would come to the same thing and be even clearer (to my eyes).

src/video-player/components/runtime.tsx
```diff
@@ -67,14 +72,21 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
...
             label: "English",
             src: authoredState.captionUrl,
-            "default": true
+            default: captionsOnByDefault
           }, false);
+          // Store a ref to the captions track s we can check visibility later. There is no easy toggle event for captions show/hide
+          captionsTrackRef.current = player.textTracks()[0];
+          if (captionsTrackRef.current) {
+            captionsTrackRef.current.addEventListener("cuechange", (cue) => {
+              log("cue change", { videoUrl: authoredState.videoUrl });
```
sfentress, line 87: Would it be worth including what the cue change event actually was (which cue is now visible/whatever)?

PR #45, opened 2020-07-23
Update to `@concord-consortium/slate-editor` 0.5.0

kswenson:
- required some CSS changes for the image dialog

@mklewandowski I realize you haven't worked much with this repository, but the code involved is more `slate-editor` than `question-interactives` so it'll be more familiar to you than anyone else. The styles of the `slate-editor` dialogs are not as encapsulated as they probably should be, so it requires some non-trivial CSS synchronization at this point.

### DIFF
--- a/src/shared/widgets/rich-text/rich-text-widget.scss
+++ b/src/shared/widgets/rich-text/rich-text-widget.scss
@@ -1,14 +1,38 @@
 .customRichTextWidget {
 
-  :global .modal-dialog {
+  :global .ccrte-modal-dialog {
     margin: 0;
 
     .dialog {
+      top: 0;
+      left: 50%;
+      transform: translate(-50%, 0);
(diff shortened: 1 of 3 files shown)


src/shared/widgets/rich-text/rich-text-widget.scss
```diff
@@ -1,14 +1,38 @@
 .customRichTextWidget {
 
-  :global .modal-dialog {
+  :global .ccrte-modal-dialog {
     margin: 0;
 
     .dialog {
+      top: 0;
```
kswenson, line 7: In the authoring environment, we can be in a fairly small scrollable iframe, so the top of the frame is often the best we can do.

mklewandowski: This makes sense.

PR #46, opened 2020-07-28
Image question

dstrawberrygirl:
This PR incorporates #173381241 for Image Question Display, and #173202524 for the Authoring view. This does not include snapshot or upload functionality yet, and Reporting view is to be addressed separately.


scytacki:
This has been open for a while. 
@sfentress do you have time to review it soon? Or is there another plan for this work, that isn't clear from this PR.

PR #47, opened 2020-07-28
RTE converts empty paragraphs to empty strings on export

kswenson:
[[#174055769]](https://www.pivotaltracker.com/n/projects/2441242/stories/174055769)


kswenson:
It's part of the RTE's core data model because it's necessary to support multi-line text. You could get away without it for single-line text, but we don't know whether it's single-line or multi-line until the user's done editing. In theory, you could strip the outer `<p>` tags for single-line text on export and then put it back on import, but I don't know that it'd be worth it. I think the existing TinyMCE-based editor works the same way.

PR #54, opened 2020-09-25
Add support for modal in Activity Player to the Image Interactive

dstrawberrygirl:
As per pair session with Matt, this adds a "am I modal?" option (isShowingModal) to the image so that the component can render in a different way when shown modal. In this case, we can show the higher-resolution image (if provided) when in modal state. Since the Activity Player is managing the state, this has to be passed from the parent.


mklewandowski (approved):
👍 This is working well in the activity player!  I would vote that we merge this in so we can build test activities in the Activity Player that point at a proper library interactive URL and then put in a PR in the Activity Player project.

PR #55, opened 2020-09-28
Remove on-screen track of last viewed timestamp

dstrawberrygirl:
A quick fix for the visible 0 seen on screen if the student has never watched the video before.


pjanik (approved):
The test is gone, but assuming it's a quick fix and you're in a different project now, it seems fine to me.

PR #56, opened 2020-09-29
Add `linkedInteractives` support, necessary hooks, and use them in ImageQuestion

pjanik:
LARA provides `linkedInteractives` in a separate array within initMessage. This PR adds two new hooks that let question interactives handle that in a bit easier way and provide authoring interface:
- useLinkedInteractives
- useLinkedInteractivesAuthoring

The second commit updates ImageQuestion just to provide an example.

It all might not be super obvious, but hopefully comments help. Also, hooks are more complicated than their usage (as usual). Looking at the image-question implementation, it seems it's pretty simple to link an interactive.

@scytacki, I haven't thought about adding a custom widget or field. It could be a nice option too. The only thing is that LARA expects all the linked interactives to be provided in one array. This makes widget implementation not so great, as it'd need to monitor other linked interactive widgets too, collect all the values, and generate one array before sending it to LARA. It'd make sense to provide a widget if we could set linked interactives independently, e.g. based on the label or something like that.

This PR depends on: https://github.com/concord-consortium/lara/pull/634

### DIFF
--- a/src/image-question/components/app.tsx
+++ b/src/image-question/components/app.tsx
@@ -6,7 +6,6 @@ import { Runtime } from "./runtime";
 import { StampCollection } from "../../drawing-tool/components/app";
 import { drawingToolAuthoringProps, stampCollectionDefinition, drawingToolAuthoringSchema } from "../../drawing-tool/components/app";
 
-
 export interface IAuthoredState extends IAuthoringInteractiveMetadata {
   version: number;
   hint?: string;
@@ -19,6 +18,8 @@ export interface IAuthoredState extends IAuthoringInteractiveMetadata {
   answerType: string;
   defaultAnswer: string;
   modalSupported?: boolean;
+  useSnapshot?: boolean;
+  snapshotTarget?: string;
 }
 
 export interface IInteractiveState extends IRuntimeInteractiveMetadata {
@@ -58,6 +59,16 @@ const baseAuthoringProps = {
       defaultAnswer: {
         type: "string",
         title: "Default answer"
+      },
+      useSnapshot: {
+        title: "Use snapshot button",
+        type: "boolean"
+      },
+      snapshotTarget: {
+        title: "Snapshot target",
+        type: "string",
+        enum: [],
+        enumNames: []
       }
     },
     dependencies: {
@@ -117,5 +128,6 @@ export const App = () => (
     Runtime={Runtime}
     baseAuthoringProps={baseAuthoringProps}
     isAnswered={isAnswered}
+    linkedInteractiveProps={[{ label: "snapshotTarget", supportsSnapshots: true }]}
   />
 );
--- a/src/shared/hooks/use-linked-interactives-authoring.test.ts
+++ b/src/shared/hooks/use-linked-interactives-authoring.test.ts
@@ -73,7 +73,7 @@ describe("useLinkedInteractives", () => {
     });
   });
 
-  it("does not calls setLinkedInteractives when it's not necessary (linkedInteractives array and authoredState are matching)", () => {
+  it("does not call setLinkedInteractives when it's not necessary (linkedInteractives array and authoredState are matching)", () => {
     initMessage = {
       mode: "authoring",
       linkedInteractives: [{id: "ID1", label: "linkedInteractive1"}]
--- a/src/shared/hooks/use-linked-interactives-authoring.test.ts
+++ b/src/shared/hooks/use-linked-interactives-authoring.test.ts
@@ -109,9 +109,16 @@ describe("useLinkedInteractives", () => {
         schema: {}
       });
     };
-    renderHook(HookWrapper);
+    const { rerender } = renderHook(HookWrapper);
+    expect(setLinkedInteractivesMock).not.toHaveBeenCalled(); // Nothing should happen during initial render!
+    rerender();
+    expect(setLinkedInteractivesMock).not.toHaveBeenCalled(); // Still nothing, authored state wasn't updated.
+    authoredState = {
+      linkedInteractive1: "ID2"
+    };
+    rerender();
     expect(setLinkedInteractivesMock).toHaveBeenCalledWith({
-      linkedInteractives: [{id: "ID1", label: "linkedInteractive1"}]
+      linkedInteractives: [{id: "ID2", label: "linkedInteractive1"}]
     });
   });
 
@@ -121,7 +128,7 @@ describe("useLinkedInteractives", () => {
       linkedInteractives: [{id: "ID1", label: "linkedInteractive1"}]
     };
     authoredState = {
-      linkedInteractive1: undefined
+      linkedInteractive1: "ID2"
     };
     const HookWrapper = () => {
       return useLinkedInteractivesAuthoring({
@@ -131,7 +138,14 @@ describe("useLinkedInteractives", () => {
         schema: {}
       });
     };
-    renderHook(HookWrapper);
+    const { rerender } = renderHook(HookWrapper);
+    expect(setLinkedInteractivesMock).not.toHaveBeenCalled(); // Nothing should happen during initial render!
+    rerender();
+    expect(setLinkedInteractivesMock).not.toHaveBeenCalled(); // Still nothing, authored state wasn't updated.
+    authoredState = {
+      linkedInteractive1: undefined
+    };
+    rerender();
     expect(setLinkedInteractivesMock).toHaveBeenCalledWith({
       linkedInteractives: []
     });
@@ -153,9 +167,16 @@ describe("useLinkedInteractives", () => {
         schema: {}
       });
     };
-    renderHook(HookWrapper);
+    const { rerender } = renderHook(HookWrapper);
+    expect(setLinkedInteractivesMock).not.toHaveBeenCalled(); // Nothing should happen during initial render!
+    rerender();
+    expect(setLinkedInteractivesMock).not.toHaveBeenCalled(); // Still nothing, authored state wasn't updated.
+    authoredState = {
+      linkedInteractive1: "new ID 2"
+    };
+    rerender();
     expect(setLinkedInteractivesMock).toHaveBeenCalledWith({
-      linkedInteractives: [{id: "new ID", label: "linkedInteractive1"}]
+      linkedInteractives: [{id: "new ID 2", label: "linkedInteractive1"}]
     });
   });
 });
--- a/src/shared/hooks/use-linked-interactives-authoring.test.ts
+++ b/src/shared/hooks/use-linked-interactives-authoring.test.ts
@@ -93,7 +93,7 @@ describe("useLinkedInteractives", () => {
     expect(setLinkedInteractivesMock).not.toHaveBeenCalled();
   });
 
-  it("monitors authoredState updates and calls setLinkedInteractives when necessary - 1", () => {
+  it("monitors authoredState updates and calls setLinkedInteractives when necessary - 1", async () => {
     initMessage = {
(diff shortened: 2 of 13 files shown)


src/shared/hooks/use-linked-interactives-authoring.tsx
```diff
@@ -0,0 +1,106 @@
...
+        } else if (linkedInteractive && linkedInteractive.id !== authoredStateVal) {
+          // Update array item.
+          const idx = currentLinkedInteractives.indexOf(linkedInteractive);
+          const newArray = currentLinkedInteractives.slice();
+          newArray.splice(idx, 1, {
+            id: authoredStateVal,
+            label: name
+          });
+          setCachedLinkedInteractives(newArray); // Set cached value
+          setLinkedInteractives({linkedInteractives: newArray}); // Send to LARA
```
pjanik, line 65: There's a bug here, these calls should be outside the loop, and only one array should be generated. It won't be visible in practice, as an author could change only one field at the time, but still.

src/shared/hooks/use-linked-interactives-authoring.test.ts
```diff
@@ -0,0 +1,161 @@
...
+          }
+          // Note that linkedInteractive3 wasn't listed in the original schema, so it's not added here either.
+        }
+      });
+      expect(getInteractiveListMock.mock.calls[0][0]).toEqual({ scope: "page", supportsSnapshots: true });
+      expect(getInteractiveListMock.mock.calls[1][0]).toEqual({ scope: "page", supportsSnapshots: undefined });
+    });
+  });
+
+  it("does not calls setLinkedInteractives when it's not necessary (linkedInteractives array and authoredState are matching)", () => {
```
kswenson, line 76:
```suggestion
  it("does not call setLinkedInteractives when it's not necessary (linkedInteractives array and authoredState are matching)", () => {
```

src/shared/hooks/use-linked-interactives-authoring.tsx
```diff
@@ -0,0 +1,106 @@
...
+  label: string;
+  supportsSnapshots?: boolean;
+}
+
+export interface IProps {
+  linkedInteractiveProps?: ILinkedInteractiveProp[];
+  schema: JSONSchema6;
+}
+
+type AuthoredState = {[key: string]: any};
```
kswenson, line 18: I believe `Record<string, unknown>` or `Record<string, any>` is the recommended way to type an arbitrary object at this point.

src/shared/hooks/use-linked-interactives-authoring.tsx
```diff
@@ -0,0 +1,106 @@
...
+          setCachedLinkedInteractives(newArray); // Set cached value
+          setLinkedInteractives({linkedInteractives: newArray}); // Send to LARA
+        } else if (linkedInteractive && authoredStateVal === undefined) {
+          // Remove item from the array.
+          const idx = currentLinkedInteractives.indexOf(linkedInteractive);
+          const newArray = currentLinkedInteractives.slice();
+          newArray.splice(idx, 1);
+          setCachedLinkedInteractives(newArray); // Set cached value
+          setLinkedInteractives({linkedInteractives: newArray}); // Send to LARA
+        } else if (linkedInteractive && linkedInteractive.id !== authoredStateVal) {
```
kswenson, line 56:
```suggestion
        } else if (linkedInteractive?.id !== authoredStateVal) {
```

src/shared/hooks/use-linked-interactives-authoring.tsx
kswenson: Perhaps add a note in the code in case the situation comes up in the future.

src/shared/hooks/use-linked-interactives-authoring.tsx
```diff
@@ -0,0 +1,106 @@
...
+            label: name
+          });
+          setCachedLinkedInteractives(newArray); // Set cached value
+          setLinkedInteractives({linkedInteractives: newArray}); // Send to LARA
+        }
+      });
+    }
+  }, [authoredState, cachedLinkedInteractives, currentLinkedInteractives, initMessage?.mode, linkedInteractiveProps]);
+
+  // Get the interactive list that are on the same page.
```
kswenson, line 71:
```suggestion
  // Get the list of interactives that are on the same page.
```

src/shared/hooks/use-linked-interactives.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    };
+    renderHook(HookWrapper);
+    expect(setAuthoredState).toHaveBeenCalledTimes(2);
+    const updatedState1 = setAuthoredState.mock.calls[0][0](authoredState);
+    const updatedState2 = setAuthoredState.mock.calls[1][0](updatedState1);
+    expect(updatedState1).toEqual({ linkedInteractive1: "new ID 1", linkedInteractive2: "ID 2" });
+    expect(updatedState2).toEqual({ linkedInteractive1: "new ID 1", linkedInteractive2: "new ID 2" });
+  });
+
+  it("do nothing if the mode is not authoring or runtime", () => {
```
kswenson, line 88:
```suggestion
  it("does nothing if the mode is not authoring or runtime", () => {
```

src/shared/hooks/use-linked-interactives.tsx
```diff
@@ -0,0 +1,36 @@
+import { useEffect, useRef } from "react";
+import { useAuthoredState, useInitMessage } from "@concord-consortium/lara-interactive-api";
+
+// This hook accepts list of props that are pointing to other interactives and ensures the most recent ID
```
kswenson, line 4:
```suggestion
// This hook accepts a list of props that are pointing to other interactives and ensures the most recent ID
```

src/shared/hooks/use-linked-interactives.tsx
```diff
@@ -0,0 +1,36 @@
...
+    // In the authoring mode an author might later change a property that is a linked interactive. It shouldn't
+    // be reset in this case. This case it's handled by useLinkedInteractivesAuthoring hook.
+    if (linkedInteractiveNames &&
+        authoredState &&
+        (initMessage?.mode === "authoring" || initMessage?.mode === "runtime") &&
+        !initialLinkedInteractivesProcessed.current)
+    {
+      linkedInteractiveNames.forEach(name => {
+        const authoredStateVal = authoredState[name];
+        const linkedInteractive = linkedInteractives && linkedInteractives.find(l => l.label === name);
```
kswenson, line 24:
```suggestion
        const linkedInteractive = linkedInteractives?.find(l => l.label === name);
```

src/shared/hooks/use-linked-interactives.tsx
```diff
@@ -0,0 +1,36 @@
...
+        if (!linkedInteractive && authoredStateVal !== undefined) {
+          // Linked interactive no longer present, clear the authoredState value.
+          setAuthoredState((prevAuthoredState: any) => Object.assign({}, prevAuthoredState, {[name]: undefined}));
+        } else if (linkedInteractive && linkedInteractive.id !== authoredStateVal) {
```
kswenson, line 28:
```suggestion
        } else if (linkedInteractive?.id !== authoredStateVal) {
```

kswenson (approved):
👍 Looks good -- just a handful of minor/optional suggestions

pjanik:
Thanks @kswenson. Thinking about it more, I think there's some serious design issue here:

1. On one hand, useLinkedInteractives hook will treat `linkedInteractives` array as a source of truth and overwrite authoredState properties. This is true at the interactive initialization time.
2. On the other hand, useLinkedInteractivesAuthoring hook will do the opposite - it'll treat authoredState as a source of truth and update linkedInteractives array based on it. This is true during authoring time.

Things work because of timing - 1. is executed just once at the beginning, 2. is executed usually later, only after the author makes some change to the authored state. But this doesn't seem to fit into the idea of functional programming (and good design generally). ;) 

I think I'd be good rework it, so only 1. is true - `linkedInteractives` array is always a source of truth and it overwrites authoredState properties. 

To do it, I think I'll need to implement a different approach - skip observing of authored state, and add a custom widget/field that updates linked interactives on change instead of authored state. Pretty much @scytacki's idea. But this would also require changes in LARA API and client. It should be possible to set just one linked interactive at the time (without providing the whole array to keep widget implementation sane) and linkedInteractives should be observable like authored or interactive state. 

Or we could leave it as it is, continue with this implementation (as it works), get snapshots working, and get back to the refactoring later. The good thing is that it's mostly invisible hooks refactoring, Image Question (and possibly other interactives) shouldn't change too much (they have a minimal amount of code anyway). Any opinion @scytacki? I think I'd prefer to timebox this refactoring to 1 day and see how it goes. I'm just not sure how strict are deadlines for the snapshot feature.

src/shared/hooks/use-linked-interactives-authoring.tsx
pjanik: I've fixed that problem in the recent commits.

src/shared/hooks/use-linked-interactives-authoring.tsx
pjanik: Thanks, I've updated it.

src/shared/hooks/use-linked-interactives-authoring.tsx
pjanik: This actually won't work here, as the logic of these two expressions is different.

`else if (linkedInteractive && linkedInteractive.id !== authoredStateVal)` will never enter the block if `linkedInteractive` is a falsy value.

`else if (linkedInteractive?.id !== authoredStateVal) {` might enter the block when `authoredStateVal` is equal to false or undefined.

src/shared/hooks/use-linked-interactives.tsx
pjanik: `linkedInteractives` might be equal to false (actually not undefined) and TS is complaining that .find doesn't exist on type `false | ...`.

pjanik:
I've refactored and fixed another issues here:

> On the other hand, useLinkedInteractivesAuthoring hook will do the opposite - it'll treat authoredState as a source of truth and update linkedInteractives array based on it. This is true during authoring time.

I've added one more condition - it's necessary for authoredState to change between calls and be different than `null`. That will ensure that we don't overwrite linkedInteractives on initial render/mount and won't be racing against `use-linked-interactives` hook. Test are now updated to check it.

Not sure how clear it is for others, but I'd say I'm pretty confident about these hooks now and tests should cover the important (edge) cases.

kswenson (approved):
👍 I don't claim to have fully grokked all of the complexities here, but I feel better knowing that you have. 😉

pjanik:
Tests are passing once I've updated NPM package, so I'm merging in this PR.

src/image-question/components/app.tsx
```diff
@@ -19,6 +18,8 @@ export interface IAuthoredState extends IAuthoringInteractiveMetadata {
   answerType: string;
   defaultAnswer: string;
   modalSupported?: boolean;
+  useSnapshot?: boolean;
```
scytacki, line 21: This used to be a source property instead of a boolean flag. 
At least it was in the UI: 
image

However making it a set of flags does seem more flexible. It would let an author setup a image question that lets a user take a snapshot or upload or just start drawing.

scytacki (approved):
Looks good to me. 
I know you merged it already. :)

src/image-question/components/app.tsx
pjanik: Okay, I'll keep that in mind while adding upcoming features and change this field.

PR #57, opened 2020-10-02
Update ImageQuestion and DrawingTool to support snapshots

pjanik:
[#173594575]


src/drawing-tool/components/runtime.tsx
```diff
@@ -45,10 +45,31 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
     }
   };
   // useRef to avoid passing interactiveState into useEffect, or it will reload on every drawing edit
-  const interactiveStateRef = useRef<any>(interactiveState);
-  const setInteractiveStateRef = useRef<((state: any) => void)>(handleSetInteractiveState);
-  interactiveStateRef.current = interactiveState;
-  setInteractiveStateRef.current = handleSetInteractiveState;
+  const initialInteractiveStateRef = useRef<any>(interactiveState);
+  const setinitialInteractiveStateRef = useRef<((state: any) => void)>(handleSetInteractiveState);
```
kswenson, line 49:
```suggestion
  const setInitialInteractiveStateRef = useRef<((state: any) => void)>(handleSetInteractiveState);
```
?

kswenson (approved):
👍 Looks good

src/drawing-tool/components/runtime.tsx
pjanik: Thanks, fixed.

PR #58, opened 2020-10-06
Support hostFeatures of initInteractive message

kswenson:
Replace `authoredState` injection with use of `hostFeatures` property in `initInteractive` message in image interactive.

### DIFF
--- a/src/image/components/app.tsx
+++ b/src/image/components/app.tsx
@@ -17,8 +17,7 @@ export interface IAuthoredState {
   creditLinkDisplayText?: string;
   allowLightbox?: boolean;
   scaling?: "fitWidth" | "originalDimensions";
-  modalSupported?: boolean;
-  isShowingModal?: boolean;
+  isShowingModal?: boolean; // TODO: move this to initInteractive message
 }
 
 const baseAuthoringProps = {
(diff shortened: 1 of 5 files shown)


src/image/components/app.tsx
```diff
@@ -17,8 +17,7 @@ export interface IAuthoredState {
   creditLinkDisplayText?: string;
   allowLightbox?: boolean;
   scaling?: "fitWidth" | "originalDimensions";
-  modalSupported?: boolean;
-  isShowingModal?: boolean;
+  isShowingModal?: boolean; // TODO: move this to initInteractive message
```
pjanik, line 20: So, I think I don't understand this line (also the previous ones). Why `isShowingModal` in authoredState?
Is ActivityPlayer injecting this value into authoredState when it's rendering interactive using modal?

kswenson: It's basically an ad hoc mechanism for Activity Player to communicate with the interactive. As the comment suggests, it would be better for there to be a better mechanism than using the authoredState for this purpose.

dstrawberrygirl (approved):
I think this looks good. As you mentioned, this is the slightly unglamorous technique to get modal popout versions of the same component to render in the Activity Player. This is a step in the right direction.

pjanik (approved):

PR #59, opened 2020-10-06
Move snapshot support to Drawing Tool from Image Question, refactor both interactives

pjanik:
This PT story explains why it happened: https://www.pivotaltracker.com/story/show/175127634
We had some discussion yesterday about differences between Image Question and Drawing Tool.
It's a bit blurry, but after talking to Leslie I summarized all the differences in the linked PT story.

So, I've moved Snapshot from IQ to DT and tried to simplify IQ. It doesn't take pieces of DT anymore. It mostly extends it (especially authoring schema and interfaces should be more clear).

use-linked-interactives-authoring hook now supports properties that are nested too (e.g. in dependencies part).

### DIFF
--- a/src/drawing-tool/components/app.tsx
+++ b/src/drawing-tool/components/app.tsx
@@ -4,7 +4,7 @@ import { JSONSchema6 } from "json-schema";
 import { IRuntimeInteractiveMetadata, IAuthoringInteractiveMetadata } from "@concord-consortium/lara-interactive-api";
 import { BaseQuestionApp } from "../../shared/components/base-question-app";
 
-// Note that TS interfaces should match JSON schema. Currently there's no way to generate one from the other.
+// Note that TS interfaces should match JSON schema. Currently there"s no way to generate one from the other.
 // TS interfaces are not available in runtime in contrast to JSON schema.
 
 export interface StampCollection {
@@ -16,14 +16,20 @@ export interface StampCollection {
 export interface IAuthoredState extends IAuthoringInteractiveMetadata {
   version: number;
   hint?: string;
-  backgroundImageUrl?: string;
-  imageFit: string;
-  imagePosition: string;
-  stampCollections: StampCollection[];
+  prompt?: string;
+  required?: boolean;
+  predictionFeedback?: string;
+  backgroundSource?: "url" | "upload" | "snapshot";
+  backgroundImageUrl?: string; // predefined by author
+  snapshotTarget?: string;
+  imageFit?: string;
+  imagePosition?: string;
+  stampCollections?: StampCollection[];
 }
 
 export interface IInteractiveState extends IRuntimeInteractiveMetadata {
-  drawingState: string;
+  drawingState?: string;
+  userBackgroundImageUrl?: string; // snapshot or upload done by student
 }
 
 export const stampCollectionDefinition = {
@@ -110,12 +116,7 @@ export const stampCollectionDefinition = {
   }
 };
 
-export const drawingToolAuthoringProps = {
-  backgroundImageUrl: {
-    title: "Background Image URL",
-    type: "string",
-    format: "uri"
-  },
+const backgroundProps = {
   imageFit: {
     title: "Background image fit",
     type: "string",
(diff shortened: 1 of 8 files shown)


src/drawing-tool/components/app.tsx
```diff
@@ -4,7 +4,7 @@ import { JSONSchema6 } from "json-schema";
 import { IRuntimeInteractiveMetadata, IAuthoringInteractiveMetadata } from "@concord-consortium/lara-interactive-api";
 import { BaseQuestionApp } from "../../shared/components/base-question-app";
 
-// Note that TS interfaces should match JSON schema. Currently there's no way to generate one from the other.
+// Note that TS interfaces should match JSON schema. Currently there"s no way to generate one from the other.
```
kswenson, line 7:
```suggestion
// Note that TS interfaces should match JSON schema. Currently there's no way to generate one from the other.
```

src/drawing-tool/components/app.tsx
```diff
@@ -206,10 +275,17 @@ const baseAuthoringProps = {
...
+    imageFit: {
+      "ui:widget": "radio"
+    },
+    imagePosition: {
+      "ui:widget": "radio"
+    }
   },
-
-  // Can't get defaults to work with custom stamps, so validating would be ugly and confusing for users until
+  // Can"t get defaults to work with custom stamps, so validating would be ugly and confusing for users until
```
kswenson, line 288:
```suggestion
  // Can't get defaults to work with custom stamps, so validating would be ugly and confusing for users until
```

src/drawing-tool/components/runtime.tsx
```diff
@@ -28,48 +29,51 @@ interface DrawingToolOpts {
...
+    } else if (authoredState.backgroundSource === "upload" || authoredState.backgroundSource === "snapshot") {
+      backgroundImgSrc = userBackgroundImageUrl;
+    }
+
+    const bgPosition = authoredState.imagePosition || "center";
+    const bgFit = authoredState.imageFit || "shrinkBackgroundToCanvas";
     const imageOpts = {
-      src: bgProps.backgroundImageUrl,
-      position: bgProps.imagePosition
+      src: backgroundImgSrc, // not that undefined / null is a valid value (used to remove background)
```
kswenson, line 66:
```suggestion
      src: backgroundImgSrc, // note that undefined / null is a valid value (used to remove background)
```

src/drawing-tool/components/runtime.tsx
```diff
@@ -106,41 +110,58 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, sna
...
+      {
+        useSnapshot && authoredState.snapshotTarget &&
+        <button className={cssHelpers.laraButton} onClick={handleSnapshot} disabled={snapshotInProgress} data-test="snapshot-btn">
+          <CameraIcon className={cssHelpers.smallIcon} /> { interactiveState?.userBackgroundImageUrl ? "Replace snapshot" : "Take a snapshot" }
+        </button>
+      }
+      { snapshotInProgress && <p>Please wait while the snapshot is being taken...</p> }
+      {
+        useSnapshot && authoredState.snapshotTarget === undefined &&
+        <p className={css.warn}>Snapshot won&apos;t work, as the target interactive is not selected</p>
```
kswenson, line 163:
```suggestion
        <p className={css.warn}>Snapshot won&apos;t work, as no target interactive is selected</p>
```

src/shared/components/base-authoring.tsx
```diff
@@ -46,6 +46,7 @@ export const BaseAuthoring = <IAuthoredState,>({ authoredState, setAuthoredState
 
   // This hook provides list of interactives on a given page and saving of the linked interactive IDs.
   const schemaWithInteractives = useLinkedInteractivesAuthoring({ linkedInteractiveProps, schema });
+  // const schemaWithInteractives = schema;
```
kswenson, line 49: Seems like this comment could be removed at this point.

src/image-question/components/runtime.tsx
```diff
@@ -1,33 +1,19 @@
...
 
-interface IProps extends IRuntimeQuestionComponentProps<IAuthoredState, IInteractiveState> {
-}
+interface IProps extends IRuntimeQuestionComponentProps<IAuthoredState, IInteractiveState> {}
 
 const kGlobalDefaultAnswer = "Please type your answer here.";
 
 export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, setInteractiveState, report }) => {
-  const { version, imageFit, imagePosition, required, stampCollections, modalSupported } = authoredState;
+  const { required, modalSupported } = authoredState;
```
kswenson, line 15: FYI: there will be a conflict here with #58 which removes the `modalSupported` property from authoredState.

kswenson (approved):
👍 Looks great! Just a few minor nits to pick.

src/image-question/components/runtime.tsx
pjanik: Right, I need to check how this works, for now I'm ignoring the modal version and I'm focusing on features that work inline.

pjanik:
Thanks for the comments, I've fixed all the mentioned lines, so going to merge it.

PR #61, opened 2020-10-09
Image Question, Drawing Tool and Image refactoring

pjanik:
The main goal here was to make sure that ImageQuestion uses dialog and generally follows LARA implementation. To do it, I've divided DrawingTool interactive into smaller components that are now reused by the ImageQuestion.

Also, I've removed `modalSupported` and `isShowingModal` usage, as ActivityPlayer will no longer proivde them:
https://github.com/concord-consortium/activity-player/pull/71
And changed ImageInteractive to always use "lightbox" modal type, as now it's implemented both in LARA and ActivityPlayer.


pjanik:
This PR depends on: https://github.com/concord-consortium/activity-player/pull/71

kswenson (approved):
👍 Looks good!

PR #62, opened 2020-10-12
Image Question and Drawing Tool updates

pjanik:
1. Image Question saves PNG when the user closes the dialog using "Close" button
2. Image Question and Drawing Tool layout is updated to match LARA styles
3. Image Question uses special metadata types that will let LARA and ActivityPlayer export it in the same way as the classic LARA image question. It's necessary so the answer image URL can be displayed in the reports.


src/image-question/components/runtime.tsx
```diff
@@ -18,31 +20,70 @@ const drawingToolDialogUrlParam = "drawingToolDialog";
...
     setControlsHidden(false);
-    showModal({ type: "dialog", url: window.location.href + "?" + drawingToolDialogUrlParam });
+    if (success) {
+      openDrawingToolDialog();
+    }
+  };
+
+  const openDrawingToolDialog = () => {
+    // notCloseable: true disabled click-to-close backdrop and X icon in the corner. Dialog can be closed only
+    // manually by calling closeModal() from one of our own widgets.
```
kswenson, line 39: "manually" suggests user intervention; how about "via API" instead?

src/shared/hooks/use-property-update.ts
```diff
@@ -0,0 +1,18 @@
+import usePrevious from "react-hooks-use-previous";
+import { useState } from "react";
+
+const loading = "loading";
+
+// A simple hook that accepts object and a property name and returns `true` when the property has been updated.
+// There's an assumption that the object can be initially unavailable (undefined) and this should not affect
+// the result. This hook will detect change of the property, not property and object combination.
+// It's useful for observing interactive state properties.
+export const usePropertyUpdate = (object: any, propertyName: string) => {
```
kswenson, line 10: I found the name `usePropertyUpdate` less than transparent on first reading. How would you feel about `usePropertyDidChange` or some variant thereof?

kswenson (changes requested):
👍 Looks good -- just a few minor suggestions.

src/shared/hooks/use-property-update.ts
pjanik: I renamed it, thanks. All the naming suggestions are welcome. ;)

kswenson (approved):
👍 Changes look good!

PR #63, opened 2020-10-16
Image Question: add "Make drawing" button, make drawing changes detection more reliable

pjanik:
This PR adds "Make drawing" button when no snapshot and no upload modes are used. I've forgotten about this before as I was focused on these special modes.

Also, drawing tool updates should be detected more reliably now. I've removed a hook that I used before, as it wasn't working for "Make drawing" use case.

[#175127634]


kswenson (approved):
👍 Looks good

PR #65, opened 2020-10-28
Image Question: warn users not to use external image URLs, use LARA img proxy when necessary

pjanik:
Depends on: https://github.com/concord-consortium/lara/pull/645

[#175127634]

### DIFF
--- a/src/drawing-tool/components/drawing-tool.tsx
+++ b/src/drawing-tool/components/drawing-tool.tsx
@@ -63,6 +63,9 @@ export const DrawingTool: React.FC<IProps> = ({ authoredState, interactiveState,
         const url = new URL(backgroundImgSrc || "");
         if (!url.host.match(/concord\.org/)) {
           // Use LARA image proxy to avoid tainting canvas when external image URL is used.
+          // Note that concord.org domain is actually not enough when the subdomains are different.
+          // So, the host needs to have CORS headers enabled. It's more likely for servers owned by CC. In practice
+          // it should be mostly AWS S3 bucket with a concord.org domain.
           backgroundImgSrc = LARA_IMAGE_PROXY + backgroundImgSrc;
         }
       } catch (e) {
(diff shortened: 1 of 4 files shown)


src/drawing-tool/components/drawing-tool.tsx
```diff
@@ -56,6 +59,15 @@ export const DrawingTool: React.FC<IProps> = ({ authoredState, interactiveState,
     let backgroundImgSrc: string | undefined;
     if (authoredState.backgroundSource === "url") {
       backgroundImgSrc = authoredState.backgroundImageUrl;
+      try {
+        const url = new URL(backgroundImgSrc || "");
+        if (!url.host.match(/concord\.org/)) {
+          // Use LARA image proxy to avoid tainting canvas when external image URL is used.
```
scytacki, line 65: I think it is also required that the image needs to have CORS headers. Not every site at CC will have those.
I think the approach above is still OK, but I think there should be a comment here about the CORS headers being needed. 

If there are no CORS headers, what happens? Will the drawing not save, or will there be an error when the background is first loaded?

scytacki (changes requested):
Looks good. Just one request to add a comment.

src/drawing-tool/components/drawing-tool.tsx
pjanik: > I think it is also required that the image needs to have CORS headers. Not every site at CC will have those.

Right, I was assuming that if we test against concord.org, we'll allow sites like models-resources.concord.org and similar. But this is not 100% correct of course. Just a compromise to avoid using proxy all the time.

> If there are no CORS headers, what happens? Will the drawing not save, or will there be an error when the background is first loaded?

This is handled by the Drawing Tool library. First, it'll try to request image with crossOrigin=anonymous. This request will fail if the server doesn't have CORS headers. In this case, it'll print a warining in the console and request an image again without crossOrigin headers. The background will be set and visible for users, but saving will fail.

Drawing Tool could fail even earlier, after the first request. But author / user would still need to open the image question dialog to see that something is broken. And it seems it's not really a concern of Drawing Tool app, as it doesn't really know if anyone would like to use `.toDataUrl()` (although I think we're the only users of this lib, so we could adjust it a bit).

We could also add some more advanced validation to authoring interface. I think that react-json-schema-form allows async validation. In this case we could make a request even in the authoring UI.

scytacki (approved):
OK this seems good enough to me. If authors run into it a lot, then we can consider authoring UX.

PR #66, opened 2020-10-30
174426725 question interactive carousel

emcelroy:
Related PT stories:

https://www.pivotaltracker.com/story/show/174426725
https://www.pivotaltracker.com/story/show/175150323

[#174426725]
[#175150323]

The carousel interactive is essentially a clone of the scaffolded question interactive with some changes. That means there is some duplicate code. If we end up trying to consolidate things, that will likely be done later on in a separate PR.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,8 @@
   },
   "scripts": {
     "start": "webpack-dev-server --inline --hot --content-base dist/",
+    "start:secure": "webpack-dev-server --https --inline --hot --content-base dist/ --cert ./LocalhostCertificates/localhost.pem --key ./LocalhostCertificates/localhost.key",
+    "start:secure:no-certs": "webpack-dev-server --https --inline --hot --content-base dist/",
     "build": "npm-run-all lint:build clean build:webpack",
     "build:webpack": "webpack --mode production --devtool false",
     "clean": "rimraf dist",
@@ -134,6 +136,7 @@
     "react-dom": "^16.13.1",
     "react-hooks-use-previous": "^1.1.1",
     "react-jsonschema-form": "^1.8.1",
+    "react-responsive-carousel": "^3.2.10",
     "resize-observer-polyfill": "^1.5.1",
     "shutterbug": "^1.3.0",
     "striptags": "^3.1.1",
--- a/src/carousel/components/iframe-authoring.tsx
+++ b/src/carousel/components/iframe-authoring.tsx
@@ -1,0 +1,110 @@
+import React, { ChangeEvent, useEffect, useRef, useState, useCallback } from "react";
+import { FieldProps } from "react-jsonschema-form";
+import { IframePhone } from "../../shared/types";
+import iframePhone from "iframe-phone";
+import deepEqual from "deep-equal";
+
+import css from "./iframe-authoring.scss";
+import { v4 as uuidv4 } from "uuid";
+
+// This is only temporary list. In the future, it will be replaced by LARA Interactive API call that returns all the available managed interactives.
+const carouselQuestionSegment = /carousel\/?$/;
+const availableInteractives = [
+  {
+    url: "",
+    name: "Select an interactive"
+  },
+  {
+    url: window.location.href.replace(carouselQuestionSegment, "open-response"),
+    name: "Open response"
+  },
+  {
+    url: window.location.href.replace(carouselQuestionSegment, "fill-in-the-blank"),
+    name: "Fill in the blank"
+  },
+  {
+    url: window.location.href.replace(carouselQuestionSegment, "multiple-choice"),
+    name: "Multiple choice"
+  },
+  {
+    url: window.location.href.replace(carouselQuestionSegment, "video-player"),
+    name: "Video"
+  }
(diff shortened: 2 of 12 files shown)


package.json
```diff
@@ -35,6 +35,8 @@
   },
   "scripts": {
     "start": "webpack-dev-server --inline --hot --content-base dist/",
+    "start:secure": "webpack-dev-server --https --inline --hot --content-base dist/ --cert ./LocalhostCertificates/localhost.pem --key ./LocalhostCertificates/localhost.key",
+    "start:secure:no-certs": "webpack-dev-server --https --inline --hot --content-base dist/",
```
pjanik, line 39: Out of curiosity - did you have to use the version with certs? I usually use a command similar to the no-certs variant.

src/carousel/components/iframe-authoring.tsx
```diff
@@ -0,0 +1,110 @@
...
+  },
+  {
+    url: window.location.href.replace(carouselQuestionSegment, "multiple-choice"),
+    name: "Multiple choice"
+  },
+  {
+    url: window.location.href.replace(carouselQuestionSegment, "video-player"),
+    name: "Video"
+  }
+];
```
pjanik, line 33: This will cause the same bug as Scaffolded Question one: https://www.pivotaltracker.com/story/show/175502689
But it's fine to leave it like that. I'll update SQ soon, so then we can copy the same approach here.

src/carousel/components/runtime.tsx
```diff
@@ -0,0 +1,109 @@
...
+  };
+
+  // TODO: Instead of the below, make a pause API in lara-interactive-api
+  const pauseAllVideos = () => {
+    const iframes = document.querySelectorAll("iframe");
+    iframes.forEach( (iframe: any) => {
+      const video = iframe.contentWindow.document.querySelector("video");
+      video?.pause();
+    });
+  };
```
pjanik, line 68: This is a bit brutal but should work in practice before we have a proper API. :) Did you check if video-interactive recognizes that the video was paused in a programmatic way and updates its controls and state correctly?

pjanik (commented):
I think my biggest concern is about meeting this requirement:

> In Teacher Reports, each interactive will be treated as it's own question regardless of display.
(https://www.pivotaltracker.com/n/projects/2441242/stories/174426725)

The current approach / carousel would be treated just as one question in all the reports, and it won't be easy to change that. If we really need questions to be reported separately, it seems that it should be some layout option supported by ActivityPlayer. Or make a bunch of changes to somehow report subquestions independently, but this might not be easy and very clear long-term. @scytacki would probably know if this requirement is important and what's the plan here.

Another thing is that I don't see the report mode implemented. But maybe it's going to be a separate PR. It should be quite easy - just pass report property from Corousel to IframeRuntime. I bet ScaffoldedQuestion is doing that.

Otherwise, it looks good to me. There are a few inline comments but nothing major.
Code duplication seems like the biggest issue. But I agree this can be addressed later in another PR. I'd just make sure that we either work on that relatively soon or at least add PT story.

scytacki:
I'm pretty sure the teacher report requirement is not needed for this work. 
I put a comment in the story to get Leslie's feedback.

package.json
emcelroy: No, I didn't use the version with certs, but added that option in case it was useful in the future.

src/carousel/components/runtime.tsx
emcelroy: @pjanik I didn't check that, at least not completely. It's not immediately clear to me how to check the state, but I'll look into it. If you have any suggestions about how to check that, please let me know. For what it's worth, when looking at the carousel in runtime in a web browser, the videos do pause when they should and their controls behave just as you would expect them to when paused.

pjanik: > For what it's worth, when looking at the carousel in runtime in a web browser, the videos do pause when they should and their controls behave just as you would expect them to when paused.

That's nice, and I meant mostly that. Actually video interactive saves its state (e.g. progress) if I remember correctly, but I'm not sure how well that works in practice. One additional test would be to use a carousel so the video is paused, wait a bit, reload a page and see if the progress has been saved correctly (if video progress matches the moment when you used the carousel). But first I'd check if it works like that even without a carousel. ;)

emcelroy: Yes, I tested standalone video interactives and they do save their progress. If you pause a video, reload the page, then press the video's play button, it starts off where you had paused it before.

The videos within the carousel behave the same way, even if they're paused using the brutal method :) So that's good.

One semi-related issue is that the carousel doesn't save its state. So if you're on the second item and reload the page, the carousel returns to the first item. I'm not sure what the desired behavior is there, but I'll ask.

scytacki (approved):
Looks good to me.

scytacki:
I think all of @pjanik's comments were addressed. 
There might be a need for more work which passes the report mode from the carousel into the sub interactives, but perhaps even without that the report view will be good enough for now. So that can be a separate PR if it is needed.

PR #67, opened 2020-10-30
Support authored state migrations, fix Scaffolded Question

pjanik:
https://www.pivotaltracker.com/story/show/175502689

@scytacki, I'm marking you as a reviewer here, as we briefly discussed the issue and possible solutions before.

This fix became a bit lengthy, as I've decided to add a concept of authored state migration. I could kept using `url` property and just treat it as a path segment, but I've tried to make that a bit more clear and hopefully add a tool that can be used by other interactives later.

Re migrations - one issue with the current implementation is that when a random component decides to use `useAuthoredState`, it might get outdated state. So, it'd be its responsibility to call migrations first. It shouldn't be a problem right now as all interactive-specific components use authored state provided via props, but something to keep in mind. We could extend `lara-interactive-api` and register migration function there. Then, the client could apply migration globally, so we wouldn't have to care about it while using `useAuthoredState` hook. But I'm not yet sure if it's worth it yet.

Re Scaffolded Question issue - `libraryInteractiveId` name replaces old `url` property of the subinteractive. Currently, these IDs are hardcoded ("open-response", "multiple-choice", "fill-in-the-blank"), but in the future, we might use real IDs provided by LARA API. There's also a function that converts this ID to a real URL. The current implementation uses Scaffolded Question URL to generate subquestion URL, but in the future this could be replaced by LARA API too. I've decided to use this naming, so we don't have to make more migrations in the future here. And it kinda matches the current implementation too. 

If we ever decide to use LARA API and full interactive library, we wouldn't have to make another authoredState migration. We will need to migrate `libraryInteractiveId` values in LARA database though ("open-response" to open response library interactive ID etc.), but it seems this can't be avoided anyway.

### DIFF
--- a/cypress/integration/scaffolded-question.test.js
+++ b/cypress/integration/scaffolded-question.test.js
@@ -13,13 +13,13 @@ context("Test scaffolded question interactive", () => {
       phonePost("initInteractive", {
         mode: "runtime",
         authoredState: {
-          version: 1,
+          version: 2,
           prompt: "Test prompt",
           hint: "Hint",
           subinteractives: [
             {
               id: "int1",
-              url: "/open-response",
+              libraryInteractiveId: "open-response",
               authoredState: {
                 version: 1,
                 prompt: "Subquestion prompt",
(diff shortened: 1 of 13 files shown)


scytacki (approved):
Looks great to me.

PR #68, opened 2020-11-19
Graph interactive

pjanik:
Demo:
https://lara-master.concordqa.org/activities/20038/pages/304665/edit - open preview in Activity Player.
I can either add tests to this PR when they're ready or open a new PR. Just thought some feedback earlier could be useful too.

### DIFF
--- a/src/graph/components/app.tsx
+++ b/src/graph/components/app.tsx
@@ -1,0 +1,38 @@
+import React from "react";
+import { BaseApp } from "../../shared/components/base-app";
+import { IAuthoredState } from "./types";
+import { Runtime } from "./runtime";
+import { JSONSchema6 } from "json-schema";
+
+const baseAuthoringProps = {
+  schema: {
+    type: "object",
+    properties: {
+      version: {
+        type: "number",
+        default: 1
+      },
+      dataSourceInteractive: {
+        title: "Data Source Interactive",
+        type: "string",
+        enum: [],
+        enumNames: []
+      }
+    }
+  } as JSONSchema6,
+
+  uiSchema: {
+    version: {
+      "ui:widget": "hidden"
+    }
+  }
+};
+
+export const App = () => (
+  <BaseApp<IAuthoredState>
+    Runtime={Runtime}
+    baseAuthoringProps={baseAuthoringProps}
+    disableAutoHeight={false}
+    linkedInteractiveProps={[{ label: "dataSourceInteractive" }]}
+  />
+);
--- a/src/graph/components/app.tsx
+++ b/src/graph/components/app.tsx
@@ -12,11 +12,38 @@ const baseAuthoringProps = {
         type: "number",
         default: 1
       },
-      dataSourceInteractive: {
-        title: "Data Source Interactive",
+      dataSourceInteractive1: {
+        title: "Data Source Interactive 1",
         type: "string",
         enum: [],
         enumNames: []
+      }
+    },
+    // Why this strange chain of dependencies?
+    // Theoretically we could use dataSourceInteractive.type === "array", as react-jsonschema-forms handles that nicely.
+    // But it would make handling of linked interactives difficult. Our hooks / helpers support only top-level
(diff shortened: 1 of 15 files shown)


src/graph/components/app.tsx
```diff
@@ -0,0 +1,67 @@
...
+        type: "string",
+        enum: [],
+        enumNames: []
+      }
+    },
+    // Why this strange chain of dependencies?
+    // Theoretically we could use dataSourceInteractive.type === "array", as react-jsonschema-forms handles that nicely.
+    // But it would make handling of linked interactives difficult. Our hooks / helpers support only top-level
+    // linked interactive properties in the authored state. That way authored state will remain "flat", having just:
+    // .dataSourceInteractive1, ..., .dataSourceInteractive5 properties instead of an array.
```
kswenson, line 26: Comment says "5" but we only seem to go up to "3".

src/graph/components/runtime.tsx
```diff
@@ -0,0 +1,85 @@
...
+        const newDataset = newLinkedIntState && newLinkedIntState.dataset;
+        // Accept null or undefined datasets too to clear them. If it's an object, make sure it follows
+        // specified format.
+        if (!newDataset || newDataset && newDataset.type === "dataset" && Number(newDataset.version) === 1) {
+          setDatasets(prevDatasets => {
+            const newDatasets = prevDatasets.slice();
+            newDatasets[datasetIdx] = newDataset;
+            return newDatasets;
+          });
+        } else if (newDataset && newDataset.type === "dataset" && Number(newDataset.version) !== 1) {
```
kswenson, line 53: Repeating the expected version number in several places in these tests seems sub-optimal. Perhaps consolidate the test:
```typescript
  const isValidDatasetVersion = Number(newDataset.version) === 1;
```

src/graph/generate-chart-data.ts
```diff
@@ -0,0 +1,136 @@
+import { ChartData, ChartDataSets } from "chart.js";
+import { IDataset } from "@concord-consortium/lara-interactive-api";
+
+// Make Chart.js types a bit more specific (remove ? from properties definitions), so it's easier to work with them.
```
kswenson, line 4: If all you're doing is making optional properties required, consider using [`Required<T>`](https://mariusschulz.com/blog/mapped-type-modifiers-in-typescript#removing-the-mapped-type-modifier).

src/graph/components/runtime.tsx
```diff
@@ -0,0 +1,85 @@
...
+export interface IProps {
+  authoredState: IAuthoredState;
+}
+
+const defaultGraphOptions = {
+  scales: {
+    yAxes: [
+      {
+        ticks: {
+          beginAtZero: true,
```
kswenson, line 18: Is this necessarily the right default? Shouldn't it depend on the data? It could be a reasonable default for an empty graph, so maybe it belongs below.

kswenson (approved):
👍 Looks good -- just a few comments/suggestions

src/graph/components/app.tsx
pjanik: Thanks, fixed.

src/graph/components/runtime.tsx
pjanik: Done.

src/graph/generate-chart-data.ts
pjanik: Unfortunately Required makes all the properties required. I need only a few.

src/graph/components/runtime.tsx
pjanik: Without this option, the Y axis is pretty random (based on current values) and two bar graphs with very different values can look identical. I think it can be pretty confusing for students. Screenshot when this option is disabled:
Screenshot 2020-11-20 at 12 48 15
Note that all the bars have the same height even though values are 2, 3, and 4. 
@lbondaryk, any opinion on that?

kswenson: I think I was imagining scatter plots when I made the comment. For bar charts I think you're right it is the appropriate default.

PR #70, opened 2020-11-24
Authoring/Display custom buttons on carousel

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/175724522

[#175724522]

* Add custom image for nav button option to carousel

### DIFF
--- a/src/carousel/components/app.tsx
+++ b/src/carousel/components/app.tsx
@@ -43,6 +43,9 @@ const baseAuthoringProps = {
             },
             authoredState: {
               type: "any"
+            },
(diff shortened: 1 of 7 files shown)


scytacki (commented):
This seems fine to me. But I haven't looked at closely. I defer to @pjanik.

pjanik (approved):
Looks good to me too. 👍 I've tested that in practice, as I was surprised it doesn't reload subquestion authoring when you change the nav image URL, but then I realized there's an additional check in the `useEffect` that prevents it.

PR #71, opened 2020-11-24
Add graph interactive Jest and Cypress tests, update Cypress to 6.0

pjanik:
[#173954114]

This PR adds missing tests. 

Cypress tests might not be super intuitive, as they include implementation details related to linked interactives handling, listenerIds, interactive list etc., but the benefit is that we have these functionalities covered now too.

I've updated Cypress to 6.0.0 as it didn't cause any issues.


kswenson (approved):
👍 Looks good!

PR #72, opened 2020-11-25
Update graph interactive to better handle missing xAxisProp in datasets

pjanik:
[#175876890]

https://www.pivotaltracker.com/story/show/175876890

xAxisProp can be undefined now. Row indices will be displayed then. I've also added displayXAxisLabels as it might be useful to hide them in some cases (Vortex has datasets just with one row).

### DIFF
--- a/src/graph/generate-chart-data.ts
+++ b/src/graph/generate-chart-data.ts
@@ -57,11 +57,15 @@ export const generateChartData = (datasets: Array<IDataset | null | undefined>) 
     if (!dataset) {
       return;
     }
-    const xPropIdx = dataset.properties.indexOf(dataset.xAxisProp);
-    if (xPropIdx < 0) {
-      throw new Error("Incorrect dataset.xAxisProp value");
-    }
-    const labels = dataset.rows.map(row => row[xPropIdx] || "");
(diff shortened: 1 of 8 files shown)


src/graph/generate-chart-data.ts
```diff
@@ -57,11 +57,15 @@ export const generateChartData = (datasets: Array<IDataset | null | undefined>)
...
-    }
-    const labels = dataset.rows.map(row => row[xPropIdx] || "");
+    const xPropIdx = dataset.xAxisProp ? dataset.properties.indexOf(dataset.xAxisProp) : -1;
+    const labels: Label[] = dataset.rows.map((row, rowIdx) => {
+      if (xPropIdx === -1) {
+        // If x property isn't defined, use row index, but starting from 1.
+        return rowIdx + 1;
+      }
+      const val = row[xPropIdx];
+      return val !== null ? val : "";
```
kswenson, line 67: Assuming that `undefined` is not a valid value either, I would use `!= null` here to catch `undefined` as well.

kswenson (approved):
👍 LGTM

src/graph/generate-chart-data.ts
pjanik: `undefined` is not allowed by the interface (because of possible JSON serialization along the way), so it should never show up here.

PR #75, opened 2020-12-10
Image Q and Drawing Tool: display warning if bg image isn't CORS enabled

pjanik:
[#175167437]

@scytacki, implementation follows what we discussed.

It's displayed in runtime, but it means it'll be also displayed in authoring preview. So, I think it should be pretty clear. Image question will display an error message BEFORE the dialog is open, so authors are less likely to miss this error. Screenshots below. Let me know if I should use some different wording.

Screenshot 2020-12-10 at 19 55 21
Screenshot 2020-12-10 at 19 55 26


src/drawing-tool/components/runtime.tsx
```diff
@@ -17,6 +19,14 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
   const readOnly = report || (authoredState.required && interactiveState?.submitted);
   const useSnapshot = authoredState?.backgroundSource === "snapshot";
   const useUpload = authoredState?.backgroundSource === "upload";
+  const authoredBgCorsError = useCorsImageErrorCheck({
+    performTest: authoredState.backgroundSource === "url" && !!authoredState.backgroundImageUrl,
+    imgSrc: authoredState.backgroundImageUrl
+  });
+  if (authoredBgCorsError) {
+    return <div className={css.error}>Authored background image is not CORS enabled, please use a different background
```
kswenson, line 27:
```suggestion
    return <div className={css.error}>Authored background image is not CORS enabled; please use a different background
```
Either a semicolon or two full sentences to avoid the run-on sentence.

src/shared/hooks/use-cors-image-error-check.ts
```diff
@@ -0,0 +1,43 @@
...
+      // it should be mostly AWS S3 bucket with a concord.org domain.
+      return true;
+    }
+  } catch (e) {
+    // Malformed URL, ignore it.
+  }
+  return false;
+};
+
+export const useCorsImageErrorCheck = (props: { performTest: boolean, imgSrc?: string }): boolean => {
```
kswenson, line 30:
```suggestion
export const useCorsImageErrorCheck = ({ performTest, imgSrc }: { performTest: boolean, imgSrc?: string }): boolean => {
```
if you fully destructure here you should be able to eliminate all reference to `props` below.

kswenson (approved):
👍 -- just a couple minor suggestions

src/shared/hooks/use-cors-image-error-check.ts
pjanik: Thanks, fixed.

src/drawing-tool/components/runtime.tsx
pjanik: Thanks, fixed.

PR #77, opened 2020-12-16
Drag and drop interactive

pjanik:
A very basic version of drag and drop interactive.

An author can specify canvas dimensions and a list of draggable objects.
They'll be positioned on the left side of the canvas automatically. Student can move them around.

@sfentress, I'm marking you as a reviewer as it uses a similar drang and drop approach as DESE Silt model.


sfentress (approved):
This looks good to me.

PR #78, opened 2020-12-18
Add drag and drop interactive tests

pjanik:
[#176013709]

I was trying to get Cypress working with react-dnd but I haven't managed to do it yet. I might try again on Monday if I have some time between other tasks.


sfentress (approved):
Looks great. :+1:

PR #81, opened 2021-01-13
Add text decoration to fill in the blank and image question types

mklewandowski:
This PR adds text decoration to the fill in the blank and image question types.  This was reported as a bug by @kevinsantos-cc when reviewing the glossary work.  Adding text decoration to these question types appears to simply have been an oversight when the initial glossary work was completed.

### DIFF
--- a/src/image-question/components/runtime.tsx
+++ b/src/image-question/components/runtime.tsx
@@ -108,7 +108,11 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
     const inlineImage = interactiveState?.answerImageUrl || interactiveState?.userBackgroundImageUrl || authoredBackgroundUrl;
     // Render answer prompt and answer text in inline mode to replicate LARA's Image Question UI
     return <div>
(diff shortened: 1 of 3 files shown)


src/image-question/components/runtime.tsx
```diff
@@ -105,7 +108,7 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
     const inlineImage = interactiveState?.answerImageUrl || interactiveState?.userBackgroundImageUrl || authoredBackgroundUrl;
     // Render answer prompt and answer text in inline mode to replicate LARA's Image Question UI
     return <div>
-      { authoredState.prompt && <div>{renderHTML(authoredState.prompt)}</div> }
+      { authoredState.prompt && <DecorateChildren decorateOptions={decorateOptions}><div>{renderHTML(authoredState.prompt)}</div></DecorateChildren> }
```
kswenson, line 111: I would break this out into multiple lines given how much is going on in this one line at this point.

kswenson (approved):
👍 Looks good

src/image-question/components/runtime.tsx
mklewandowski: Good call.   Looking at that again, it is pretty ugly on one line.  Fixed.

PR #82, opened 2021-01-14
172877789 te mc wrapper

mklewandowski:
This PR allows the MC question types to receive overlay show/hide custom messages and in turn show or hide the correct and distractors overlays.  These messages are sent from the Teacher Edition plugin when a wrapper tab is selected.

### DIFF
--- a/src/multiple-choice/components/runtime.tsx
+++ b/src/multiple-choice/components/runtime.tsx
@@ -119,7 +119,6 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
   const readOnly = report || (authoredState.required && interactiveState?.submitted);
 
   const renderRadioChecks = () => {
-    console.log(authoredState);
     return authoredState.choices && authoredState.choices.map(choice => {
       const checked = selectedChoiceIds.indexOf(choice.id) !== -1;
       const inputId = baseElementId + choice.id;
--- a/src/multiple-choice/components/runtime.tsx
+++ b/src/multiple-choice/components/runtime.tsx
@@ -44,15 +44,19 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
   const isScorable = !!authoredState.choices && authoredState.choices.filter(c => c.correct).length > 0;
 
   useCustomMessages((msg: ICustomMessage) => {
-    if (msg.type === "teacher-edition:showCorrectOverlay") {
-      setShowCorrect(true);
-      setShowDistractor(false);
-    } else if (msg.type === "teacher-edition:showDistractorOverlay") {
-      setShowDistractor(true);
-      setShowCorrect(false);
-    } else if (msg.type === "teacher-edition:hideOverlay") {
-      setShowCorrect(false);
-      setShowDistractor(false);
+    switch (msg.type) {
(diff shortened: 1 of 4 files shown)


src/multiple-choice/components/runtime.tsx
```diff
@@ -117,6 +134,16 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
             readOnly={readOnly}
             disabled={readOnly}
           />
+          {choice.correct !== undefined && choice.correct === true && showCorrect &&
```
kswenson, line 137:
```suggestion
          {choice.correct && showCorrect &&
```
The test against `undefined` is redundant (if it's `true` it can't be `undefined`). Furthermore, unless there's some need to distinguish truthy values other than `true`, the simple boolean test should be sufficient.

src/multiple-choice/components/runtime.tsx
```diff
@@ -117,6 +134,16 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
...
+            <div className={css.markContainer}>
+              <CheckMarkIcon className={`${css.mark} ${css.correctCheck}`} />
+            </div>
+          }
+          {choice.correct !== undefined && choice.correct === false && showDistractor &&
```
kswenson, line 142:
```suggestion
          {!choice.correct && showDistractor &&
```
As above, the test against `undefined` is redundant (if it's `false` it can't be `undefined`). The specific test for `false` may still be necessary if other falsy values (e.g. `undefined`) should _not_ be marked, but if that's not the case I would opt for the simpler test.

kswenson (changes requested):
👍 Looks good but I would simplify a couple of the tests as described inline.

src/multiple-choice/components/runtime.tsx
mklewandowski: Good call.  That was verbose.  Fixed.

src/multiple-choice/components/runtime.tsx
mklewandowski: Same as above.  Thanks!  Fixed.

src/multiple-choice/components/runtime.tsx
```diff
@@ -39,6 +43,19 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
...
+    if (msg.type === "teacher-edition:showCorrectOverlay") {
+      setShowCorrect(true);
+      setShowDistractor(false);
+    } else if (msg.type === "teacher-edition:showDistractorOverlay") {
+      setShowDistractor(true);
+      setShowCorrect(false);
+    } else if (msg.type === "teacher-edition:hideOverlay") {
+      setShowCorrect(false);
+      setShowDistractor(false);
+    }
```
kswenson, line 56: With three branches I'd probably opt for a switch statement, but it's a matter of preference.

kswenson (approved):
👍 Looks good

src/multiple-choice/components/runtime.tsx
mklewandowski: I switched to a switch (that sounds odd when said out loud).

scytacki (approved):
Looks good to me.

PR #83, opened 2021-02-01
"Alt text" for carousel button images

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/176575201

[#176575201]

The carousel allows authors to specify images to be used as buttons in the carousel navbar. The images are displayed by making them a background image for the corresponding `button` elements. 

These changes add the option to customize text that appears to screen readers for these buttons, as well as on mouse over of the buttons. The original PT story calls for "alt text" but since the buttons aren't actually images, we add the text as button text and as `title` and `aria-label` attributes.

### DIFF
--- a/src/carousel/components/iframe-authoring.tsx
+++ b/src/carousel/components/iframe-authoring.tsx
@@ -38,7 +38,7 @@ const availableInteractives = [
 
 export const IframeAuthoring: React.FC<FieldProps> = props => {
   const { onChange, formData } = props;
-  const { url, authoredState, id, navImageUrl } = formData;
+  const { url, authoredState, id, navImageUrl, navImageAltText } = formData;
   const [ iframeHeight, setIframeHeight ] = useState(300);
   const [ authoringOpened, setAuthoringOpened ] = useState(false);
   const iframeRef = useRef<HTMLIFrameElement>(null);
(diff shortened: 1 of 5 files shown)


src/carousel/components/iframe-authoring.tsx
```diff
@@ -113,7 +142,10 @@ export const IframeAuthoring: React.FC<FieldProps> = props => {
             <div className={css.navButtonField}>
               <label htmlFor={navImageUrl}>Custom Navigation Button Image URL</label>
               <input className="form-control" type="text" id={navImageUrl} defaultValue={navImageUrl} onChange={handleNavImageUrlChange} />
-              <p className="help-block">To customize the button for this slide, specify an image URL. Optimal image size: 150x100 pixels.</p>
+              <p className="help-block">To customize the button for this slide, enter an image URL. Optimal image size: 150x100 pixels.</p>
+              <label htmlFor={navImageAltText}>Custom Navigation Button Image Alt Text</label>
+              <input className="form-control" type="text" id={navImageAltText} defaultValue={navImageAltText} onChange={handleNavImageAltTextChange} />
```
pjanik, line 147: I think you might want to replace `{` with `"` in `htmlFor={navImageAltText}` and `id={navImageAltText}`. 
So, it's `htmlFor="navImageAltText"` and `id="navImageAltText"`.
Right now, this ID will change dynamically when the user is typing.

pjanik (approved):
Looks good. There's one inline comment that I'd fix. The current version will work too, but it's a bit strange and can generate some warnings if IDs are non-unique or don't follow DOM standards (if there are any for IDs).

src/carousel/components/iframe-authoring.tsx
emcelroy: Ah, right. Thanks for pointing that out. It's fixed in the latest commit.

PR #86, opened 2021-02-01
Fix LARA to ActivityPlayer conversion script.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/176612039

[#176612039]

This updates the full name of the image embeddable type to match what's on production.


scytacki (approved):
Seems fine to me.

PR #87, opened 2021-02-02
Fix conversion script view link

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/176767044

[#176767044]

The success message for a conversion provided a "View" link to a LARA version of the activity instead of an Activity Player version. Since it seemed like linking to the latter would require hardcoding some domains into the script, I decided to simply remove the "View" link. There is still the "Edit" link which authors can follow and then preview the activity in the Activity Player from authoring.

These changes also set the new resource's runtime to "Activity Player."


scytacki (approved):
looks good to me

PR #88, opened 2021-02-02
Carousel subquestion response issues

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/176001600

[#176001600]

In some cases, prompts for fill-in-the-blank questions in a carousel were appearing in the response area of other questions in the carousel. These changes set the `subState` value within the iteration through each subinteractive. That ensures the subState is correctly set and then used for each interactive's state.

I believe this change also fixed the bug described here: https://www.pivotaltracker.com/story/show/176013162

### DIFF
--- a/cypress/integration/carousel.test.js
+++ b/cypress/integration/carousel.test.js
@@ -83,7 +83,7 @@ context("Test carousel interactive", () => {
               answerText: "Test subquestion answer"
             }
           },
-          answerText: "[Level: 1] Test subquestion answer"
+          answerText: "Test subquestion answer"
         });
       });
     });
@@ -126,7 +126,6 @@ context("Test carousel interactive", () => {
             target_id: '',
             target_name: '',
             target_value: '',
-            scaffolded_question_level: 1,
             subinteractive_url: "open-response",
             subinteractive_type: "open_response",
             subinteractive_sub_type: undefined,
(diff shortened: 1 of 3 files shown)


pjanik (changes requested):
Looks better, more bulletproof, and simpler.
Although, the previous approach should work fine theoretically. So, I think there's some hidden bug there that would be nice to fix. 

Looking at the code in master:
https://github.com/concord-consortium/question-interactives/blob/master/src/carousel/components/runtime.tsx
The previous version was trying to set subState using `currentSubinteractiveId`, but it was actually never updating this property. So, I'm not sure how this was working at all.

I think we should either save this property (assuming the carousel should remember at which slide the user has left?) or we should remove [all the references](https://github.com/concord-consortium/question-interactives/blob/113f7a138e6cdd3dfcf6fb8f3c13f69e5538fba4/src/carousel/components/runtime.tsx#L55-L59) to this property (+ update interface) not to create confusion. The current version is closer to the latter.

emcelroy:
@pjanik I had another try at getting the previous approach to work, but couldn't make it happen. I'm missing something, but can't figure it out. I went with the other option and removed the unnecessary code that you pointed out. Thanks for catching that. 

Given time restraints I think for now it's best to have this fix in place and revisit the question of whether the carousel needs to remember which slide the user left off at later on.

pjanik:
Okay, thanks. Actually, I didn't mean that one or the other solution is better.
The only thing is whether the carousel should save its currently visible interactive, but I guess it can be a separate feature if necessary. And actually, this could work fine with the current code too (that gets subState in the interation).

pjanik (approved):

PR #89, opened 2021-02-05
Question interactives need Activity Player styles

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/176517471

[#176517471]

These changes make question interactives' fonts and button styles match the basic Activity Player specs by default. Previously, question interactives matched the base LARA runtime styles by default. (The LARA style rules are still available as mixins in src/shared/styles/helpers.scss in case we decide to make question interactives support multiple styles).


pjanik (approved):
Looks good 👍

PR #92, opened 2021-02-17
Explicitly set focus on buttons when they are clicked.

emcelroy:
This is a fix for a bug found while testing changes for this PT story: https://www.pivotaltracker.com/story/show/176662615

[#176662615]

Basically, some browsers do not set focus on buttons when they are clicked. These changes explicitly set focus to clicked buttons in the carousel navbar so there is a consistent behavior -- visually highlighting the clicked button -- across all browsers.


pjanik (approved):
Looks good 👍

PR #93, opened 2021-03-29
Add trailing slashes on interactive URLS

knowuh:
The Activity Player Offline wasn't able to load library interactives (specifically the scaffolded question) without a trailing slash.

[#176963163]
PT: https://www.pivotaltracker.com/story/show/176963163


scytacki (approved):
Looks good. I just added 2 more interactives so it matches what the carousel had before.

PR #94, opened 2021-04-16
Fix improper handling of image questions in convert-old-lara script

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177800027

[#177800027]

* Set a backgroundSource value in authoredState for those image questions that need a value.

* Set linked_interactives value for snapshot image questions.


pjanik (approved):
Thanks, it looks good to me. 👍 
I'm assuming you've tested this script locally and it fixes image questions.

emcelroy:
@pjanik Thank you. Yes, I tested this locally and it fixed the problem with converting image questions.

PR #95, opened 2021-04-26
Fix conversion doesn't update runtime for converted sequence activities

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177916895

[#177916895]

This change ensures that activities within a sequence that is converted to run in the Activity Player have their runtime set to "Activity Player."


pjanik (approved):
Looks good to me 👍🏻

PR #96, opened 2021-04-27
Remove multiple choice question default name when converting activities

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/177932956

[#177932956]


emcelroy:
@scytacki Does this seem like a reasonable solution to the problem described in the [related PT story](https://www.pivotaltracker.com/story/show/177932956)?

scytacki (approved):
Seems fine to me.

Perhaps we want to remove this default text from LARA authoring in the future, but lets leave it for now.

PR #97, opened 2021-05-30
Adds label, index, value, and unit as properties of a draggable item.

eireland:
Adds a label, a value, and a unit label to each draggable item in authoring. 
Image size is also now authorable.

### DIFF
--- a/src/drag-and-drop/components/app.tsx
+++ b/src/drag-and-drop/components/app.tsx
@@ -67,6 +67,14 @@ export const baseAuthoringProps = {
               type: "string",
               format: "uri"
             },
+            imageWidth: {
+              title: "Image width",
+              type: "number"
+            },
+            imageHeight: {
+              title:"Image height",
+              type: "number"
+            },
             itemLabel: {
               title: "Image label",
               type: "string"
--- a/src/drag-and-drop/components/app.tsx
+++ b/src/drag-and-drop/components/app.tsx
@@ -94,10 +94,6 @@ export const baseAuthoringProps = {
           }
         }
       },
-      enableCheckAnswer: {
-        title: "Allow users to check answers",
-        type: "boolean",
-      },
       initialState: {
         type: "object",
         properties: {
--- a/src/drag-and-drop/components/app.tsx
+++ b/src/drag-and-drop/components/app.tsx
@@ -79,18 +79,6 @@ export const baseAuthoringProps = {
               title: "Label",
               type: "string"
             },
-            index: {
-              title: "Index",
-              type: "number"
-            },
-            itemValue: {
-              title: "Value",
-              type: "number"
-            },
-            targetMatch: {
-              title: "Target match:",
-              type: "number"
-            }
           }
         }
       },
--- a/src/drag-and-drop/components/app.tsx
+++ b/src/drag-and-drop/components/app.tsx
@@ -66,6 +66,22 @@ export const baseAuthoringProps = {
               title: "Image URL",
               type: "string",
               format: "uri"
+            },
+            itemLabel: {
+              title: "Image label",
+              type: "string"
+            },
+            index: {
+              title: "Index",
+              type: "number"
+            },
+            itemValue: {
+              title: "Value",
+              type: "number"
+            },
+            itemUnit: {
+              title: "Unit",
+              type: "string"
             }
           }
         }
--- a/src/drag-and-drop/components/app.tsx
+++ b/src/drag-and-drop/components/app.tsx
@@ -75,10 +75,6 @@ export const baseAuthoringProps = {
               title:"Image height",
               type: "number"
             },
-            itemLabel: {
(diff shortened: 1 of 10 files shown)


src/drag-and-drop/components/app.tsx
```diff
@@ -66,10 +66,38 @@ export const baseAuthoringProps = {
...
+              title: "Index",
+              type: "number"
+            },
+            itemValue: {
+              title: "Value",
+              type: "number"
+            },
+            targetMatch: {
+              title: "Target match:",
+              type: "number"
```
pjanik, line 92: What is index, itemValue and targetMatch? 
These values seem to be unused. I bet they're related to some upcoming features? Generally, I would avoiding merging that in this PR, as if the plans or schedule change, we might end up with some unused code that might be confusing. Or you might end up implementing that differently. 😉

pjanik (changes requested):
Looks good. I've left one comment about unused properties. Otherwise looks good.

src/drag-and-drop/components/app.tsx
```diff
@@ -66,10 +66,38 @@ export const baseAuthoringProps = {
...
             }
           }
         }
       },
+      enableCheckAnswer: {
+        title: "Allow users to check answers",
+        type: "boolean",
+      },
```
pjanik, line 100: Same here, seems to be unused.

src/drag-and-drop/components/app.tsx
```diff
@@ -66,10 +66,38 @@ export const baseAuthoringProps = {
...
               format: "uri"
+            },
+            imageWidth: {
+              title: "Image width",
+              type: "number"
+            },
+            imageHeight: {
+              title:"Image height",
+              type: "number"
+            },
```
pjanik, line 77: Why do we need now image width and height? It worked without it. I think it might be annoying for authors to provide pixel perfect dimensions for each image they use.

pjanik (changes requested):
Ah, one more thing. Can we remove image width and height? Or, if it was specifically requested to be able to overwrite default dimensions, I'd make them optional.

src/drag-and-drop/components/app.tsx
eireland: Index, itemValue, and targetMatch are for upcoming features. They were specified to be in this story https://www.pivotaltracker.com/story/show/178247192
ItemValue and targetMatch will be exported to the graph question interactive.

src/drag-and-drop/components/app.tsx
eireland: It's an optionaly porperty if somebody needed to adjust their image to fit in the container. I ran into that problem when I was trying to find a water droplet to test with.

src/drag-and-drop/components/app.tsx
eireland: This was specified in the mockup here https://www.pivotaltracker.com/story/show/178218427. I'll take it out and put it in the PR for that story.

src/drag-and-drop/components/app.tsx
pjanik: Okay, so can we remove it from this PR and add them when they're actually used? 
Otherwise, we can end up with some unused confusing properties if the schedule or requirements change.

pjanik:
@eireland, I wasn't sure if you're working on that or not. But before we merge, can we just remove unused properties and interfaces? That won't be any work to re-add them again when they're necessary.

src/drag-and-drop/components/types.ts
```diff
@@ -5,6 +5,12 @@ export type ItemId = string;
 export interface IDraggableItem {
   id: ItemId;
   imageUrl?: string;
+  imageWidth: number;
+  imageHeight: number;
+  itemLabel: string;
+  index: number;
+  itemValue: number;
+  targetMatch: number;
```
pjanik, line 13: Unused props still here.

src/drag-and-drop/components/draggable-item.test.tsx
```diff
@@ -4,7 +4,7 @@ import { DraggableItem } from "./draggable-item";
 
 describe("DraggableItem", () => {
   it("renders image", () => {
-    const wrapper = shallow(<DraggableItem item={{ id: "1", imageUrl: "https://image/1" }} />);
+    const wrapper = shallow(<DraggableItem item={{ id: "1", imageUrl: "https://image/1", imageWidth: 25, imageHeight: 25, itemLabel: "image label", index: 1, itemValue: 5, targetMatch: 3 }} />);
```
pjanik, line 7: Unused props still here.

pjanik (approved):
Thanks. There are a few more places with unused props. Feel free to clean it up after merging if that's easier for you, I don't want to hold this PR.

pjanik (approved):
Looks good, thanks

PR #98, opened 2021-06-08
178218427 authoring targets

eireland:
First part of adding drop targets to question interactive.
Drop targets are authorable, and draggable items can be "dropped" onto a target.
Current behavior is when an item is dropped onto a target, the item disappears from view.


pjanik (commented):
Generally looks good. 
Do you have some demo activity somewhere in LARA staging that I could play with?
Do we really need to make targetWidth and targetHeight required? I guess it'd simplify authoring if they could just use native image dimensions automatically (code should already support that).

src/drag-and-drop/components/container.tsx
```diff
@@ -63,7 +86,20 @@ export const Container: React.FC<IProps> = ({ authoredState, interactiveState, s
         };
       }
     });
-  }, [authoredState.draggableItems]);
+    authoredState.dropZones?.forEach(target => {
+      if (target.imageUrl) {
+        const img = document.createElement("img");
+        img.src = target.imageUrl;
+        img.onload = () => {
+          setTargetDimensions(prevHash => ({...prevHash, [target.id]: {width: target.targetWidth || img.width, height: target.targetHeight || img.height }}));
```
mklewandowski, line 94: based on spacing you probably want:
`setTargetDimensions(prevHash => ({...prevHash, [target.id]: {width: target.targetWidth || img.width, height: target.targetHeight || img.height}}));`

src/drag-and-drop/components/container.tsx
```diff
@@ -63,7 +86,20 @@ export const Container: React.FC<IProps> = ({ authoredState, interactiveState, s
...
+        };
+      } else {
+        setTargetDimensions(prevHash => (
+          {...prevHash, [target.id]: {width: target.targetWidth || 100, height: target.targetHeight  || 100 }}
```
mklewandowski, line 98: `{...prevHash, [target.id]: {width: target.targetWidth || 100, height: target.targetHeight  || 100}}`

src/drag-and-drop/components/container.tsx
```diff
@@ -82,22 +122,46 @@ export const Container: React.FC<IProps> = ({ authoredState, interactiveState, s
...
+    if (setInteractiveState) {
+      // Runtime mode.
+      setInteractiveState(prevState => ({
+        ...prevState,
+        answerType: "interactive_state",
+        itemTargetIds: {
+          ...prevState?.itemTargetIds,
+          [draggableItem.item.id]: targetId
+        },
+
```
mklewandowski, line 141: probably can remove this line break

src/drag-and-drop/components/container.tsx
```diff
@@ -82,22 +122,46 @@ export const Container: React.FC<IProps> = ({ authoredState, interactiveState, s
...
+      }));
+    }
+  },[setInteractiveState]);
```
mklewandowski, line 144: `}, [setInteractiveState]);`

src/drag-and-drop/components/container.tsx
```diff
@@ -82,22 +122,46 @@ export const Container: React.FC<IProps> = ({ authoredState, interactiveState, s
...
+        return;
+      } else {
+        const delta = monitor.getDifferenceFromInitialOffset() as {
+          x: number
+          y: number
+        };
+        const left = Math.round(wrapper.position.left + delta.x);
+        const top = Math.round(wrapper.position.top + delta.y);
+        moveDraggableItem(wrapper.item.id, left, top);
     }
```
mklewandowski, line 160: Is there an alignment issue here?  Does this need to be tabbed in?

src/drag-and-drop/components/container.tsx
```diff
@@ -124,11 +213,10 @@ export const Container: React.FC<IProps> = ({ authoredState, interactiveState, s
               top: Math.min(canvasHeight - margin, top)
             };
           }
-          return <DraggableItemWrapper key={item.id} item={item} position={position} draggable={!readOnly} />;
+          return targetId === undefined && <DraggableItemWrapper key={item.id} item={item} position={position} draggable={!readOnly} />;
```
mklewandowski, line 216: Might be clearer as `!targetId &&...`

src/drag-and-drop/components/draggable-item-wrapper.test.tsx
```diff
@@ -9,7 +9,7 @@ describe("DraggableItemWrapper", () => {
   it("renders DraggableItem", () => {
     const wrapper = mount(
       <DndProvider backend={TouchBackend} options={{enableMouseEvents: true}} >
-        <DraggableItemWrapper item={{ id: "1", imageUrl: "https://image/1" }} position={{ left: 1, top: 1}} draggable={true} />
+        <DraggableItemWrapper item={{ id: "1", imageUrl: "https://image/1", imageWidth: 25, imageHeight: 25}} position={{ left: 1, top: 1}} draggable={true} />
```
mklewandowski, line 12: You might break the props onto multiple lines for readability and then check spacing on the objects.
```
<DraggableItemWrapper 
  item={{id: "1", imageUrl: "https://image/1", imageWidth: 25, imageHeight: 25}} 
  position={{left: 1, top: 1}} 
  draggable={true} 
/>

File: src/drag-and-drop/components/draggable-item-wrapper.tsx [hunk 8]
```diff
@@ -27,19 +28,18 @@ export const DraggableItemWrapper: React.FC<IProps> = ({ item, position, draggab
...
-    <div
-      ref={draggable ? drag : undefined}
-      className={`${css.draggableItemWrapper} ${draggable ? css.draggable : ""}`}
-      style={position}
-      data-cy="draggable-item-wrapper"
-    >
-      <DraggableItem item={item} />
-    </div>
+    <>
+    { isDragging ? <DraggableItemPreview /> :
```
mklewandowski, line 33: Might read clearer with the ternary formatted as such:

```
{ isDragging 
  ? <DraggableItemPreview />
  : <div>
```

src/drag-and-drop/components/drop-zone-preview.scss
```diff
@@ -0,0 +1,12 @@
+@import "../../shared/styles/helpers";
+
+$border-width: 4px;
+.dropZonePreview {
+  border: $border-width dotted #144dec;
```
mklewandowski, line 5: Do we define color constants somewhere in this project?

src/drag-and-drop/components/drop-zone-wrapper.scss
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  img {
+    pointer-events: none;
+  }
+  .targetLabel {
+    position: absolute;
+    bottom: 0;
+  }
+}
```
mklewandowski, line 33: newline

src/drag-and-drop/components/drop-zone.scss
```diff
@@ -0,0 +1,14 @@
...
+  cursor: move;
+
+  img {
+    pointer-events: none;
+  }
+  .targetLabel {
+    position: absolute;
+    bottom: 0;
+  }
+}
```
mklewandowski, line 14: newline

src/drag-and-drop/components/drop-zone-preview.scss
```diff
@@ -0,0 +1,12 @@
...
+  box-shadow: 0 0 12px rgba(0,0,0,0.25);
+  transform: translate(-$border-width, -$border-width);
```
mklewandowski, line 7: This is interesting.  What was this for?  Was this part of what was causing that problem with the border outline and the position?

src/drag-and-drop/components/draggable-item-wrapper.tsx
```diff
@@ -27,19 +28,18 @@ export const DraggableItemWrapper: React.FC<IProps> = ({ item, position, draggab
     })
   });
 
-  if (isDragging) {
-    // Hide source element on dragging and render preview only.
-    return null;
-  }
```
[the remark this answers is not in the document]
eireland: At Matt's suggestion. To simplify the logic of when to show the preview vs. the DraggableItem, there is now logic in starting in line 34 to show DraggableItemPreview if isDragging, else show the DraggableItem.

src/drag-and-drop/components/drop-zone.tsx
```diff
@@ -0,0 +1,20 @@
...
+}
+
+// Just an image for now, but it can be extended in the future to include some text and other elements.
+export const DropZone: React.FC<IProps> = ({ target, highlight }) => {
+  const style={width: target.targetWidth, height: target.targetHeight};
+  return (
+    <div className={highlight? css.highlight: ""} style={style}>
+      {target.imageUrl && <img src={target.imageUrl} alt="drag target" width={target.targetWidth} height={target.targetHeight}/>}
+      <div>{target.targetLabel}
+      </div>
```
mklewandowski, line 17: This reads a little weird I would write the div on one line:
<div>{target.targetLabel}</div>

src/drag-and-drop/components/draggable-item-wrapper.tsx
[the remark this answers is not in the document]
mklewandowski: If I recall, there was a bug when dragging the drop targets where multiple previews were showing simultaneously.  This was a quick suggestion to get it working.  Instead of returning, let this component render both the target and the drag preview (I used this pattern in leafpack).  But on reviewing things I do see that it breaks the pattern used in other components.  I wonder if this would work if put back.

src/drag-and-drop/components/drop-zone-wrapper.tsx
```diff
@@ -0,0 +1,69 @@
...
+      canDrop: monitor.canDrop(),
+      getItem: monitor.getItem(),
+    }),
+  });
+
+  const zoneStyle = { left: position.left, top: position.top };
+  const highlight = isOver && canDrop;
+  return (
+    <>
+      {isDragging ? <DropZonePreview /> :
```
mklewandowski, line 52: I think this will read better as:
```
{ isDragging
 ? <DropZonePreview />
 : <div>
 ```

File: src/drag-and-drop/components/drop-zone-wrapper.tsx [hunk 16]
```diff
@@ -0,0 +1,69 @@
...
+        <div
+          ref={draggable ? drag : drop}
+          className={`${css.dropZoneWrapper}
+                      ${draggable ? css.draggable : ""}
+                      ${target.imageUrl ? "" : css.background }
+                      ${(isOver && canDrop)? css.highlight : ""}
+                      `
+                    }
```
mklewandowski, line 60: Might be easier to read as:
```
className={`${css.dropZoneWrapper}
                     ${draggable ? css.draggable : ""}
                     ${target.imageUrl ? "" : css.background }
                     ${(isOver && canDrop)? css.highlight : ""}`}
                     ```

**Review by mklewandowski — CHANGES_REQUESTED** [review]
I had some style requests, but it generally looks good.  I think some of the names of the drag and drop components could be better, but honestly these are tough things to name (and I was struggling to come up with anything much better (Maybe DropTarget instead of DropZone?).  Same question as Piotr, is there an activity where I can use this in action?  I wonder if some of the commits should be squashed?

File: src/drag-and-drop/components/container.tsx (hunk 1 above)
Comment (mklewandowski) [reply src/drag-and-drop/components/container.tsx:94]: In other words, make sure spacing around brackets is the same.
this:
`{ foo: pizza }`
or this:
`{foo: pizza}`

But not these:
`{foo: pizza }`
`{ foo: pizza}`

File: src/drag-and-drop/components/container.tsx (hunk 2 above)
Comment (mklewandowski) [reply src/drag-and-drop/components/container.tsx:98]: make sure spacing around brackets is the same

File: src/drag-and-drop/components/drop-zone-preview.scss (hunk 12 above)
Comment (pjanik) [reply src/drag-and-drop/components/drop-zone-preview.scss:7]: I guess @eireland based this on my older styles for regular item preview.
I think this transform rule was ensuring that preview with border matched the position of the actual item with border.
Otherwise, the element was shifted by these 4 pixels and there was a little image position jump on drag start.

File: src/drag-and-drop/components/drop-zone-preview.scss (hunk 9 above)
Comment (eireland) [reply src/drag-and-drop/components/drop-zone-preview.scss:5]: No that I've seen

File: src/drag-and-drop/components/drop-zone-preview.scss (hunk 12 above)
Comment (eireland) [reply src/drag-and-drop/components/drop-zone-preview.scss:7]: I took it out but can be put it back in if we want the jump.

Comment (mklewandowski) [reply src/drag-and-drop/components/drop-zone-preview.scss:7]: I would leave in whatever looks best.  I was just curious about the code.

File: src/drag-and-drop/components/container.tsx (hunk 1 above)
Comment (mklewandowski) [reply src/drag-and-drop/components/container.tsx]: And a note here, you can either have the space at the beginning/end or not have the space.  Doesn't matter as long as it is consistent.  I would follow whatever is most common in the project.

File: src/drag-and-drop/components/container.tsx (hunk 2 above)
Comment (mklewandowski) [reply src/drag-and-drop/components/container.tsx]: And a note here, you can either have the space at the beginning/end or not have the space.  Doesn't matter as long as it is consistent.  I would follow whatever is most common in the project.

File: src/drag-and-drop/components/drop-zone-preview.scss (hunk 9 above)
Comment (mklewandowski) [reply src/drag-and-drop/components/drop-zone-preview.scss:5]: No worries.  Just curious.

**Comment by mklewandowski** [comment]
Just to summarize from the grooming discussion:
- I wouldn't worry about making the dropped items do anything yet.  In the PR description you say: "Current behavior is when an item is dropped onto a target, the item disappears from view."  I think this is good enough for this PR.
- Is there a link somewhere to test this live?  An activity in the AP or LARA?  It would be nice to have an easy way to test this.

**Comment by eireland** [comment]
Can be tested here https://authoring.staging.concord.org/activities/21029/pages/309012/edit

**Review by mklewandowski — APPROVED** [review]
👍 I think this looks good!

**Review by pjanik — APPROVED** [review]
There are small sizing issues - when I drag around the drop zone label disappears and the box gets smaller. Not a big deal, but just mentioning it. Also, when I drop an item into the drop zone, it gets bigger and bigger, overflowing the available area. This looks a bit strange.

But I think this is still very nice progress, so if it's easier that way, I guess we could merge that PR and fix these issues in the upcoming PRs.

**Comment by mklewandowski** [comment]
I think the location of the draggable items and the size change of the drop targets is caused by relative vs. absolute positioning.  I would vote to merge this PR and get that work done in a subsequent PR.  @eireland we can look at that today if you want.

File: src/drag-and-drop/components/drop-zone-wrapper.tsx [hunk 17]
```diff
@@ -0,0 +1,71 @@
...
+    <>
+      { isDragging
+        ? <DropZonePreview />
+        : <div
+            ref={draggable ? drag : drop}
+            className={`${css.dropZoneWrapper}
+                        ${draggable ? css.draggable : ""}
+                        ${target.imageUrl ? "" : css.background }
+                        ${(isOver && canDrop)? css.highlight : ""}
+                      `}
```
mklewandowski, line 62: I would change this to not use use all the tabbing:
```
className={`${css.dropZoneWrapper} ${draggable ? css.draggable : ""} ${target.imageUrl ? "" : css.background } ${(isOver && canDrop)? css.highlight : ""}`}
```
Maybe break it up into some variables above.  This results in huge spaces when inspecting the DOM that makes it hard to debug:
image

src/drag-and-drop/components/draggable-item.tsx
```diff
@@ -8,10 +8,12 @@ export interface IProps {
 
 // Just an image for now, but it can be extended in the future to include some text and other elements.
 export const DraggableItem: React.FC<IProps> = ({ item }) => {
+  // console.log("item: ", item);
```
mklewandowski, line 11: remove?

src/drag-and-drop/components/drop-zone-wrapper.scss
```diff
@@ -0,0 +1,36 @@
+.dropZoneWrapper {
+  $border-width: 4px;
+
```
mklewandowski, line 3: extra line break?

src/drag-and-drop/components/drop-zone-wrapper.tsx
```diff
@@ -0,0 +1,71 @@
...
+      canDrop: monitor.canDrop(),
+      getItem: monitor.getItem(),
+      getDropResult: monitor.getDropResult(),
+    }),
+  });
+
+  const zoneStyle = {left: position.left, top: position.top};
+  const highlight = isOver && canDrop;
+  return (
+    <>
```
mklewandowski, line 53: I don't think you need the fragment.  Just remove this and remove the curly braces on the next line.

src/drag-and-drop/components/drop-zone-wrapper.tsx
kswenson: This code is begging/pleading/screaming for the use of the `classNames` library. 😉

mklewandowski:
One more note on the drop target sizing issue.  This is caused by the fact that we use `position: relative;` inside the `draggableItem` CSS class.  We can change this to `position: absolute;` and it will fix the position/sizing, but it does cause issues because then the draggable item wrapper doesn't have the right size (it becomes 0x0) and things like the highlight state won't work.  This ties to the question of if you can drag stuff out of the containers.  You could only apply the absolute position ONLY if something has been dropped which would solve the above (so draggable items that have been dropped get a CSS class added that sets the position to absolute).  However, if we end up needing to drag things out of the container, then we would have the same problem and still need a better solution.

src/drag-and-drop/components/drop-zone-wrapper.tsx
mklewandowski: I was gonna say, if Kirk was reviewing this, he would recommend `classNames` to simplify this!

mklewandowski (approved):
👍 LGTM!

PR #101, opened 2021-06-14
Removes image height and width in authoring

eireland:
Takes out draggable item image width and height from authoring


mklewandowski (approved):
👍 LGTM!

PR #102, opened 2021-06-14
Adds label, value, and target match properties to draggable items

eireland:
Remove all unused props and types


mklewandowski (approved):
👍 LGTM!

PR #104, opened 2021-06-15
Drag And Drop: add offset to dropped items

mklewandowski:
This PR adds an offset to draggable items when displayed in a drop target so that they do not overlap each other with the exact same position.  

add-drop-offset

### DIFF
--- a/src/drag-and-drop/components/draggable-item-wrapper.tsx
+++ b/src/drag-and-drop/components/draggable-item-wrapper.tsx
@@ -9,7 +9,6 @@ export interface IProps {
   item: IDraggableItem;
   position: IPosition;
   draggable: boolean;
-  isDropped?: boolean;
 }
 
 // These types are used by react-dnd.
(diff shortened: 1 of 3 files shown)


eireland (approved):
LGTM!

ddamelin:
Do we need to snap them into a position where you can no longer see what is underneath the top one? Students may decide they want to move a particular one to another drop zone, so I don't think we can cover them this way. Perhaps there is a better algorithm for placing them in the box where they won't overlap.

mklewandowski:
> Do we need to snap them into a position where you can no longer see what is underneath the top one? Students may decide they want to move a particular one to another drop zone, so I don't think we can cover them this way. Perhaps there is a better algorithm for placing them in the box where they won't overlap.

Yes, this is definitely not final.  I raised these same concerns, but for now we're just working towards an MVP before the sprint ends.  Ideally we will have a better solution for placement inside of the drop container and behavior after an item has been dropped (e.g., dropped item can be moved out of a drop container).  At present, we don't have clear designs on any of this (and hopefully there will be discussion around what we want to do since there are a variety of ways to handle this).  This PR is simply a step towards the end goal and at least allows the user to see their dropped content.

mklewandowski:
What I had recommended in another discussion was filling things in as we would a left justified flex box.  So we sort of tile the dragged content into the container starting in the upper left.  Authoring, however, allows for variable container and drag item size so we're somewhat backed into a corner for handling the edge cases.  We would need to leave it to the author to make a large enough container, allow content to spill outside the container or potentially use scroll bars if the content doesn't fit.  As you can see in the attached image, this approach works fine in a mockup where I control the container size and number of dropped items.  But what happens if the author made a small container?  What happens if they have 10 items that can be dragged to the container?  Anyway, I'll leave the design decisions to others.
image

PR #105, opened 2021-06-15
Send data to graph

eireland:
Currently only shows the sum of the values dropped onto a target.
Graph only comes up once a draggable item has been dropped.
We tried to get it to pre-load with a useEffect but it didn't work unless we put the setInteractiveState in s setTimeOut > 1000.

### DIFF
--- a/src/drag-and-drop/utils/generate-dataset.ts
+++ b/src/drag-and-drop/utils/generate-dataset.ts
@@ -12,10 +12,10 @@ export const generateDataset = ( targets: IDropZone[], targetValues?: Record<Tar
   return {
     type: "dataset",
     version: 1,
-    properties: ["body of water", "area"],
+    properties: ["x", "y"],
     // Always use first property as X axis. It might be necessary to customize that in the future, but it doesn't
     // seem useful now.
-    xAxisProp: "body of water",
+    xAxisProp: "x",
     rows: rows
   };
 };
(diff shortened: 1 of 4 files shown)


mklewandowski:
Is there somewhere to test this?

src/drag-and-drop/utils/generate-dataset.ts
```diff
@@ -0,0 +1,21 @@
...
+  return {
+    type: "dataset",
+    version: 1,
+    properties: ["x", "y"],
+    // Always use first property as X axis. It might be necessary to customize that in the future, but it doesn't
+    // seem useful now.
+    xAxisProp: "x",
+    rows: rows
+  };
+};
```
mklewandowski, line 21: Add newline

mklewandowski (approved):
👍 This looks good to me, but I'm not too familiar with this approach.  I did add one small note on improving style.  Otherwise I would wait for @pjanik to review before merging.

pjanik (approved):
There are some design issues that I didn't focus on before while I was debugging the code with @eireland, but I think this PR should be merged and we can open a new one to address that.

Our interactive state became pretty redundant, many parts of it can be derived from the other parts. E.g. targetAggreateValues could be derived from item positions / dropzone state. Similarly, the dataset could be derived from these values.

I think that storing targetAggreateValues that way will become problematic when we start moving items in and out of the dropzones. Or when we let authors change aggregate function.

My impression is that we should:
- Remove targetAggreateValues, and calculate it dynamically based on the item positons / dropzone state only when necessary.
- Dataset should be redundant and updated each time the interactive state is updated. It's not perfect, but it's a compromise done by lara-interactive-api, so sharing data is straighforward. 

But let's focus on that a bit later, probably after we also merge in @mklewandowski PR related to moving items out of the dropzones.

PR #106, opened 2021-06-16
Drag and Drop question: allow drag and drop from drop containers

mklewandowski:
This PR improves the drag and drop question so that a draggable item can be moved from one drop container to another and so a draggable item can be moved from a drop container back to the main canvas.  The implementation required storing the drop container position when the initial drop occurs.  This is required because our move code calculates the new position using a delta added to the current position.  Since the current position of an item in the drop containers is relative to the drop container, we need the drop container's position as well.  If we change the positioning or layout of items in the drop container in the future, this solution may also need to be updated.

better-drag-n-drop


pjanik (approved):
It looks good to me. I resolved some merge conflicts after I merged previous @eireland's PR.

PR #107, opened 2021-06-16
Drag and drop: refactor dataset and interactive state handling

pjanik:
This PR refactors DnD interactive state.
I've removed `targetAggregateValue` as it was redundant (see my comments here: https://github.com/concord-consortium/question-interactives/pull/105). `dataset` is now automatically calculated on each update of dependant variable (`authoredState.dropZones`, `droppedItemData` only), so it should be more reliable. Also, it automatically adds support of removing items from dropzones that was recently added by @mklewandowski.

There are two issues that are probably not related to drag and drop interactive. I've tried to add comments that explain that as clear as possible. I've tried to debug them, but it seems it can be a bigger task:
- https://www.pivotaltracker.com/story/show/178556017
- https://www.pivotaltracker.com/story/show/178556075

### DIFF
--- a/src/drag-and-drop/components/container.tsx
+++ b/src/drag-and-drop/components/container.tsx
@@ -94,6 +94,28 @@ export const Container: React.FC<IProps> = ({ authoredState, interactiveState, s
     });
   }, [authoredState.draggableItems, authoredState.dropZones]);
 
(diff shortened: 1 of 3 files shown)


eireland (approved):
Looks good to me

mklewandowski (approved):
👍 LGTM!

PR #108, opened 2021-06-16
Drag and Drop Question: Code cleanup and refactor

mklewandowski:
This PR cleans up and refactors some of the drag and drop code.  The only change to the user is that there is now a gray background on the drop target when dragging it during authoring (rather than a transparent background which looked odd).  Changes include:
- add gray background on the drop target when dragging it during authoring
- remove unused CSS classes on elements and unused CSS styles
- improve spacing
- remove unneeded react fragment
- make style imports last import
- improve comments
- remove unneeded string template literal


eireland (approved):
👍

PR #109, opened 2021-06-21
Drag and drop: save the interactive state immediately when the dataset is updated

pjanik:
Then, the linked graph interactive can receive an updated dataset immediately too. Otherwise, there's 2 seconds delay between changes in the drag and drop and an update of the graph interactive.


mklewandowski (approved):
👍 LGTM!

PR #110, opened 2021-07-01
Shows scaling button and help message

eireland:
Adds svgs for fullscreen icon
Adds fullscreen icon functionality without the embedded interactive.


src/full-screen/components/runtime.tsx
```diff
@@ -0,0 +1,31 @@
...
+      screenfull.on("change", () => {
+        setIsFullScreen(!isFullScreen);
+      });
+    }
+  };
+
+  return (
+    <div className={css.runtime}>
+      <div className={`${css.iframeContainer}`} >
+        This is the scaling interactive
```
kswenson, line 26: Presumably we don't want this text here in the production version.

src/full-screen/components/runtime.tsx
```diff
@@ -0,0 +1,31 @@
...
+      </div>
+      <FullScreenButton isFullScreen={isFullScreen} handleToggleFullScreen={onToggleFullScreen} />
```
kswenson, line 28: We probably shouldn't render the button if we don't have the ability to go fullscreen, e.g. `screenfull.isEnabled && <FullScreenButton...`.

src/full-screen/components/runtime.tsx
```diff
@@ -0,0 +1,31 @@
...
+  authoredState: IAuthoredState;
+}
+
+export const Runtime: React.FC<IProps> = ({ authoredState }) => {
+  const [ isFullScreen, setIsFullScreen ] = useState<boolean>(false);
+
+  const onToggleFullScreen = () => {
+    if (screenfull.isEnabled) {
+      screenfull.toggle();
+      screenfull.on("change", () => {
```
kswenson, line 17: We shouldn't be adding this handler on every toggle. Instead, it should be added in a `useEffect` which can then remove the handler in its cleanup function.

src/full-screen/components/runtime.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { IAuthoredState } from "./types";
+import { FullScreenButton } from "./full-screen-button";
+import css from "./runtime.scss";
+
+interface IProps {
+  authoredState: IAuthoredState;
+}
+
+export const Runtime: React.FC<IProps> = ({ authoredState }) => {
+  const [ isFullScreen, setIsFullScreen ] = useState<boolean>(false);
```
kswenson, line 12: Do we need to maintain our own state here? It seems like it could get out of sync with the actual fullscreen state given that they're being toggled independently. I'm pretty sure the fullscreen library can be queried for the current fullscreen state. If the purpose is to trigger a render when the fullscreen status changes, you can use something like this hook:
```typescript
import { useCallback, useState } from "react";

export const useForceUpdate = () => {
  const [ , setChangeCount] = useState(0);
  return useCallback(() => setChangeCount(count => count + 1), []);
};
```

src/full-screen/components/full-screen-button.tsx
```diff
@@ -0,0 +1,31 @@
...
+  });
+
+  return (
+    <div className={css.fullScreenToggle}>
+      <div id="fullScreenHelp" className={`${css.fullScreenHelp} ${isHidden? css.hidden : ""}`} >
+        Click here to enter/exit fullscreen →
+      </div>
+      <div>
+        <button className={`${css.fullScreenIcon} ${isFullScreen? css.fullscreen : ""}`}
+           onClick={handleToggleFullScreen} />
```
kswenson, line 27: Not a big deal either way, but does the `<button>` need to be inside the `<div>`?

src/full-screen/components/full-screen-button.tsx
```diff
@@ -0,0 +1,31 @@
...
+      </div>
+    </div>
+  );
+};
```
kswenson, line 31: `\n``

src/full-screen/components/runtime.scss
```diff
@@ -0,0 +1,4 @@
+@import "../../shared/styles/helpers";
+.runtime{
+  height:400px;
+}
```
kswenson, line 4: `\n`

kswenson (changes requested):
Mostly looks good -- just a few suggested tweaks.

PR #112, opened 2021-07-02
Added image uploading in authoring mode [#178284894]

dougmartin:
The new `authoring-image-upload` tool has been added to both the staging and production token-service resource list so this should work both on local Docker containers (defaults to staging), staging and production.

The final design I landed on was 1 upload folder per user.  This was we don't need to store resource ids anywhere - we just use the first `author-image-upload` resource found where the user is an owner and if none is found create one.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ npm-error.log
 .DS_Store
 .yalc
 yalc.lock
+/temp
 
 # Editors
 .idea
--- a/cypress/integration/image.test.js
+++ b/cypress/integration/image.test.js
@@ -59,9 +59,21 @@ context("Test Image interactive", () => {
 
   context("Authoring view", () => {
     it("handles pre-existing authored state", () => {
+
+      // fake token-service JWT so that image upload control is displayed above url
+      phoneListen("getFirebaseJWT", (data, phone) => {
+        phone.post("firebaseJWT", {
+          requestId: 1,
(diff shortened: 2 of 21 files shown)


dougmartin:
@scytacki See the PR description for detailed info.

This works locally with an `npm link` to the latest token-service and another `npm link` to the Lara PR referenced in the PR message.  Lara needed to be updated to enable `getFirebaseJWT` in authoring mode and the token-service 2.0 hasn't been published yet (I didn't want to go ahead and publish it in case there was a reason why it wasn't published).

scytacki:
@dougmartin did you test adding an image uploading interactive inside of a carousel and inside of a scaffolded question? 

These 'nesting' interactives need to proxy postMessage requests. It looks like they are not proxying the getFirebaseJwt messages. So I'm guessing that if you embed an interactive that has an image uploading field inside of them, it won't work.

dougmartin:
@scytacki I missed the proxying bit.  The code gracefully fails when it can't get a JWT so I didn't notice when manually testing.  I'll add it now.

scytacki (changes requested):
This looks good. 

The only issue I found was the proxying in the carousel and scaffolded question. 

Perhaps it can be taken out of Draft now that the token-service and interactive-api are released.

dougmartin:
@scytacki I've added the proxying of the firebase JWT in this commit https://github.com/concord-consortium/question-interactives/pull/112/commits/88909bc90bacb4aa038140efe38bf49e0298d0e3

This took way longer than I thought it would as I was missing the `requestId` in the proxied response so it was getting dropped by the Lara client.

scytacki (approved):
Looks good to me.

PR #114, opened 2021-07-07
172311887 wrapped interactive scaling

eireland:
Simplifies the runtime rendering by reonly rending the runtime component outside of the base question app. This was mostly to bypass a styling.
Wrapped interactive scales correctly when switching from embedded to full screen and back.


src/full-screen/components/runtime.tsx
```diff
@@ -0,0 +1,88 @@
...
+    screenfull?.on("change", onChange);
+    window.addEventListener('resize', onChange);
+    return () => {
+      screenfull?.off("change", onChange);
+      window.removeEventListener('resize', onChange);
+    };
+  }, [forceUpdate]);
+
+  const isFullScreen = screenfull?.isFullscreen;
+  const getIframeTransforms = (_window: any, _screen: any) => {
```
kswenson, line 29:
```suggestion
  const getIframeTransforms = (_window: Window, _screen: Screen) => {
```

src/full-screen/components/full-screen-button.tsx
```diff
@@ -0,0 +1,30 @@
...
+import css from "./full-screen-button.scss";
+
+interface IProps {
+  isFullScreen: boolean;
+  handleToggleFullScreen: () => void
+}
+
+export const FullScreenButton: React.FC<IProps> = (props) => {
+  const {isFullScreen, handleToggleFullScreen} = props;
+  const [isHidden, setIsHidden] = useState(false);
```
kswenson, line 12:
```suggestion
  const [isHintHidden, setIsHintHidden] = useState(false);
```
Just to make it clear that this doesn't refer to hiding the button itself.

src/full-screen/components/runtime.tsx
```diff
@@ -0,0 +1,88 @@
...
+import queryString from "query-string";
+import { FullScreenButton } from "./full-screen-button";
+import { useForceUpdate } from "../../shared/hooks/use-force-update";
+
+import css from "./runtime.scss";
+
+const screenfull = _screenfull.isEnabled ? _screenfull : undefined;
+
+// eslint-disable-next-line no-empty-pattern
+export const Runtime: React.FC = ({}) => {
```
kswenson, line 12:
```suggestion
export const Runtime: React.FC = () => {
```
then you can eliminate `// eslint-disable-next-line no-empty-pattern` on the previous line.

src/full-screen/components/runtime.tsx
```diff
@@ -0,0 +1,88 @@
...
+    const onChange = () => forceUpdate();
+    screenfull?.on("change", onChange);
+    window.addEventListener('resize', onChange);
+    return () => {
+      screenfull?.off("change", onChange);
+      window.removeEventListener('resize', onChange);
+    };
+  }, [forceUpdate]);
+
+  const isFullScreen = screenfull?.isFullscreen;
```
kswenson, line 28: I'm not sure exactly where the best place to do this would be, but I think it would be worth adding a comment about where this code came from, e.g. `This code is patterned after the jQuery-based implementation in the [Cloud File Manager](... link to CFM repo ...).`

kswenson (changes requested):
👍 Looks good -- just a few suggested tweaks.

kswenson (approved):
👍 Changes look good

PR #117, opened 2021-07-20
178846911 refactor proxy

eireland:


src/shared/components/iframe-runtime.tsx
```diff
@@ -14,31 +12,36 @@ interface ILogRequest {
...
+  id?: string;
+  iframeStyling?: any;
   interactiveState: any;
-  setInteractiveState: (state: any) => void;
-  report?: boolean;
+  logRequestData?: Record<string, unknown>;
   navImageUrl?: string;
   navImageAltText?: string;
+  report?: boolean;
+  scaffoldedQuestionLevel?: number;
```
scytacki, line 23: `scaffoldedQuestionLevel` doesn't seem to be used.

src/shared/components/iframe-runtime.tsx
```diff
@@ -14,31 +12,36 @@ interface ILogRequest {
...
-  // Why is interativeState and setInteractiveState kept in refs? So it's not necessary to declare these variables as
-  // useEffect's dependencies. Theoretically this useEffect callback is perfectly fine either way, but since
-  // it reloads the iframe each time it's called, it's not a great experience for user when that happens while he is
-  // interacting with the iframe (e.g. typing in textarea). And interactiveState is being updated very often,
-  // as well as setInteractiveState that is generated during each render of the parent component.
-  const interactiveStateRef = useRef<any>(interactiveState);
-  const setInteractiveStateRef = useRef<((state: any) => void)>(setInteractiveState);
-  interactiveStateRef.current = interactiveState;
-  setInteractiveStateRef.current = setInteractiveState;
+  ({ authoredState, id, iframeStyling, interactiveState, logRequestData, report,
```
scytacki, line 30: It looks like `navImageUrl` and `naveImageAltText` are not used here, so they could be removed from the props type definition.

src/shared/components/iframe-runtime.tsx
```diff
@@ -55,6 +58,12 @@ export const IframeRuntime: React.FC<IProps> =
       phone.addListener("hint", (newHint: IHintRequest) => {
         setHint(newHint.text || "");
```
scytacki, line 59: For the fullscreen interactive this should call setHint on the host. See my comment below where the hint is rendered.

src/shared/components/iframe-runtime.tsx
```diff
@@ -14,31 +12,36 @@ interface ILogRequest {
...
+  iframeStyling?: any;
   interactiveState: any;
-  setInteractiveState: (state: any) => void;
-  report?: boolean;
+  logRequestData?: Record<string, unknown>;
   navImageUrl?: string;
   navImageAltText?: string;
+  report?: boolean;
+  scaffoldedQuestionLevel?: number;
+  url?: string;
```
scytacki, line 24: Is there a case where the url is not set? 
It seems like the iframe would always need an url.

src/shared/components/iframe-runtime.tsx
```diff
@@ -101,13 +101,15 @@ export const IframeRuntime: React.FC<IProps> =
         phoneRef.current.disconnect();
       }
     };
-  }, [url, authoredState, report, id, navImageUrl, navImageAltText]);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  },[]);
```
scytacki, line 105: It is usually really important to useEffect to know all of its dependencies.

src/shared/types.ts
```diff
@@ -6,3 +8,40 @@ export interface IframePhone {
   initialize: () => void;
   disconnect: () => void;
 }
+
+export type WrapperType = "scaler" | "carousel" | "scaffolding";
+
+// Note that TS interfaces should match JSON schema. Currently there's no way to generate one from the other.
+// TS interfaces are not available in runtime in contrast to JSON schema.
+export interface IAuthoredState extends IAuthoringInteractiveMetadata {
```
scytacki, line 16: Each of the interactives define their own IAuthoredState and IInteractiveState types that are specific to them. 
I realize that the types for Carousel and Scaffolded question are almost identical, so it is tempting to combine them like this. But they should be kept separate for clarity and maintainability.

src/shared/components/iframe-runtime.tsx
```diff
@@ -101,13 +101,15 @@ export const IframeRuntime: React.FC<IProps> =
...
 
+  const iframeStyle = wrapper === "scaler" ? iframeStyling : {width: "100%", height: iframeHeight, border: "none"};
   return (
-    <div>
-      <iframe ref={iframeRef} src={url} width="100%" height={iframeHeight} frameBorder={0} />
+    <>
+      <iframe ref={iframeRef} src={url} style={iframeStyle} className={wrapper === "scaler" ? css.subInteractiveIframe : ""} />
       { hint &&
         <div className={css.hint}>{renderHTML(hint)}</div> }
```
scytacki, line 112: For the scaler/fullscreen interactive the hint should be set on the host instead of shown inline like this. 

So I think the best approach is to pass a `setHint` function as a property to IframeRuntime, and move this rendering of the hint up into the scaffolded question and carousel runtime components.

The scaler/fullscreen interactive can implement setHint by calling `setHint` from the interactive-api.

src/shared/components/iframe-runtime.tsx
```diff
@@ -101,13 +101,15 @@ export const IframeRuntime: React.FC<IProps> =
         phoneRef.current.disconnect();
       }
     };
-  }, [url, authoredState, report, id, navImageUrl, navImageAltText]);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  },[]);
 
+  const iframeStyle = wrapper === "scaler" ? iframeStyling : {width: "100%", height: iframeHeight, border: "none"};
```
scytacki, line 107: I'd recommend ignoring the `wrapper` here and just doing something like:
`const iframeStyle = iframeStyling ?? {width: "100%", height: iframeHeight, border: "none"}`

So that way if iframeStyling is set, then it is used, otherwise a default style is used.

Perhaps this means the scaling/fullscreen interactive needs to always be sure to send an iframeStyling prop, but as far as I can tell it is already doing this.

src/shared/components/iframe-runtime.tsx
```diff
@@ -101,13 +101,15 @@ export const IframeRuntime: React.FC<IProps> =
...
   return (
-    <div>
-      <iframe ref={iframeRef} src={url} width="100%" height={iframeHeight} frameBorder={0} />
+    <>
+      <iframe ref={iframeRef} src={url} style={iframeStyle} className={wrapper === "scaler" ? css.subInteractiveIframe : ""} />
```
scytacki, line 110: Can the `css.subInteractiveIframe` styles just be added to the `iframeStyling` passed by the scaling/fullscreen interactive?  

Then there won't be a need for this duplicate way of setting styles, and the `wrapper` property can be removed.

scytacki (changes requested):
This is looking really good. 

I found a few things to fix or improve.

src/shared/components/iframe-runtime.tsx
scytacki: looks good

src/shared/components/iframe-runtime.tsx
scytacki: looks good

scytacki (changes requested):
A minor thing to change with the url.
And a bit more work on handling hints.

src/shared/components/iframe-runtime.tsx
scytacki: @eireland you might be busy with other stuff, but I just wanted to make sure you saw these comments.

scytacki (changes requested):
As far as I can tell this PR still has an issue with setHint, see my previous comments.

src/shared/components/iframe-runtime.tsx
scytacki: looks good now.

scytacki (approved):
looks good to me.

PR #118, opened 2021-07-21
Fix multi-select MC questions set to single select on conversion.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178888327

[#178888327]

Fix incorrect property name value from `multipleAnswer` to `multipleAnswers`.


scytacki (approved):
Looks good to me.

PR #119, opened 2021-07-22
Fix videos and images in a converted activity have dimensions changed

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178888877

[#178888877]

For videos: LARA sets a default height value of 240 for "LARA native" videos. The height value doesn't seem to be used anywhere, though, and the editing form doesn't seem to allow authors to change that value, either. In LARA runtime, it seems like videos are always shown with a width of 100%. A height isn't explicitly set. So I just removed the setting of a `fixedHeight` value when a video is converted to a QI video.

For images: QI images have a `scaling` property that LARA native images don't. It seems like LARA images also always take up the full width of their containers. So I changed it so the conversion script always set an image's `scaling` property to "fitWidth".


scytacki (approved):
seems fine to me

PR #120, opened 2021-07-26
178964353 scaler aspect ratio

eireland:
Scales the wrapped iframe interactive based on screen size

### DIFF
--- a/src/full-screen/components/runtime.tsx
+++ b/src/full-screen/components/runtime.tsx
@@ -4,14 +4,14 @@ import queryString from "query-string";
 import { FullScreenButton } from "./full-screen-button";
 import { useForceUpdate } from "../../shared/hooks/use-force-update";
 import { IframeRuntime } from "../../shared/components/iframe-runtime";
-import { setHint, useInteractiveState } from "@concord-consortium/lara-interactive-api";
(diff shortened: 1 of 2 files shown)


src/full-screen/components/runtime.tsx
```diff
@@ -25,6 +25,14 @@ export const Runtime: React.FC = () => {
...
     };
   }, [forceUpdate]);
+
+  useEffect(() => {
+    const aspectRatio = screen.width/screen.height;
+    setSupportedFeatures({
+      interactiveState: true,
+      aspectRatio
+    });
+  }, [initMessage]);
```
scytacki, line 35: Did the code in the CFM try to handle the case where the screen changed sized. Perhaps if the window was dragged to second screen?  I'm guessing not, but I wanted to check since it seems like this code would not handle this case.

If the old didn't handle that case then I don't think it is necessary to handle it here either.

scytacki (changes requested):
Looks good. I just have a question about whether screen size changes are handled.

src/full-screen/components/runtime.tsx
eireland: The old one didn't handle the change in aspect ratio when changing screen size. This one does change height/aspect ratio when screen size changes.

scytacki (approved):
Sounds good to me.

PR #121, opened 2021-07-26
Video player should show first frame of video if no poster is specified

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178889194

[#178889194]

Set video preload to "metadata" so browser retrieves the video's first frame. Also make sure `getPoster` doesn't return an empty value. (If it does return an empty value, the video player will show a black rectangle instead of the first frame. That can happen if a poster value is specified for the video, but then removed.)

### DIFF
--- a/src/video-player/components/runtime.tsx
+++ b/src/video-player/components/runtime.tsx
@@ -168,7 +168,7 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
   };
 
   const getPoster = () => {
-    if (!hasStartedPlayback) {
+    if (!hasStartedPlayback && authoredState.poster !== "") {
       return authoredState.poster;
     }
   };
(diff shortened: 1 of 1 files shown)


src/video-player/components/runtime.tsx
```diff
@@ -168,7 +168,7 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
   };
 
   const getPoster = () => {
-    if (!hasStartedPlayback) {
+    if (!hasStartedPlayback && authoredState.poster !== "") {
```
dougmartin, line 171: I would change this to `if (!hasStartedPlayback && !!authoredState.poster {` in case `authoredState.poster` is `null`.

The `!!` will return `false` for both `null` and the empty string.

dougmartin (changes requested):
Change the check for the poster to account for nulls and empty strings.

src/video-player/components/runtime.tsx
emcelroy: Got it -- thanks. Change made.

scytacki (approved):
Looks good to me

PR #123, opened 2021-08-04
Fixed two issues found during QA of image uploads:

dougmartin:
- To pass the "uri" format validation spaces in the uploaded filename are converted and compressed down a single dash
- An additional check of the token-service domain was added to the image url validator


pjanik (approved):
Looks good 👍

pjanik (changes requested):
Looks good, but I think some unit tests need to be updated, as TravisCI build is failing.

dougmartin:
@pjanik the test break has been fixed in this commit: https://github.com/concord-consortium/question-interactives/pull/123/commits/0724ff85fb39bc26581b156df3a8f5d840de90d2

pjanik (approved):
Thanks, looks good.

PR #124, opened 2021-08-04
Fixed bug where navigation urls with spaces in the carousel were emitting broken css [#179119615]

dougmartin:


src/carousel/components/runtime.tsx
```diff
@@ -128,7 +129,7 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
           let buttonStyle = {};
           let buttonClass = currentSlide === index ? css.activeButton : "";
           if (interactive.navImageUrl) {
-            buttonStyle = { backgroundImage: "url(" + interactive.navImageUrl + ")" };
+            buttonStyle = { backgroundImage: cssUrlValue(interactive.navImageUrl)};
```
kswenson, line 132: It looks like `drag-and-drop` is doing something similar in container.tsx. Did you check to see whether the same fix is appropriate there?

dougmartin: Good catch.  I've updated it in this commit: https://github.com/concord-consortium/question-interactives/pull/124/commits/dbe5191b1423fcdaf8fb7c80ecc498a507334f79

knowuh (approved):
Lookin' good to me.

PR #125, opened 2021-08-16
178168627 labbook fix qa issues

knowuh:
## QA Concerns addressed:
- Target will always be "any" meaning upload and snapshot.
- When a snapshot target is specified, the snaphot button appears.
- No warning text is displayed if the snapshot target attribute is missing.
- Upload and snapshot both use normal font-weight.
- Caption textbox is constrainted to vertical resizing, and will not resize out of the parent container.

## QA Concerned not addressed:
- When you take a snapshot or upload an image, both buttons say "Please wait" - intentional?   **YES** Intentional.
- When you upload an image, it doesn't appear as a thumbnail unless you draw on it. I am not sure this is true. It does take several seconds for the thumbnail to update however.

## PT Story:
[#178168627] -- https://www.pivotaltracker.com/story/show/178168627

### DIFF
--- a/src/labbook/components/app.tsx
+++ b/src/labbook/components/app.tsx
@@ -26,6 +26,11 @@ const baseAuthoringProps = deepmerge(drawingToolBaseAuthoringProps, {
         type: "number",
         title: "How many thumbnails to display",
         default: 4
+      },
+      backgroundSource: {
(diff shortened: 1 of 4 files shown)


dougmartin (approved):
Looks good :+1:

PR #128, opened 2021-08-26
Changed the labbook question type to "iframe_interactive" to fix the Athena report generation [#178454188]

dougmartin:
Labbook was using "labbook_interactive" which caused Athena to skip adding a link url to the details report.


dougmartin:
@scytacki I've removed the `questionType` from the schema in this commit (https://github.com/concord-consortium/question-interactives/pull/128/commits/49ece7e6ef81e77922f3fb99e1279336ac00efbd) and tested it on staging and the report generates the url correctly.

scytacki (approved):
Looks good.

PR #129, opened 2021-08-31
Set answerImageUrl on sucess of save to the snapshorUrl initially created

eireland:
Currently, if a student clicks on the Take Snapshot button, which opens the draw tool with the interactive as the background image, and then closes the tab without closing the draw tool, the answer imageUrl is not set. This causes the portal report to show blank answers for the student, even though they think they have taken the snapshot.
This PR assigns answer imageUrl on click of the Take Snapshot button instead of on close of the draw tool. We will assume that the intent of the student was at least to take a snapshot of the current state of their interactive. This way there is always an answer imageUrl for the portal report.


eireland:
Added @pjanik as a possible reviewer since he created the original interactive.

pjanik (approved):
Looks good to me 👍

PR #132, opened 2021-09-08
179390297 - Fix authoring 🐞 (set background)

knowuh:
## DONE:
- fix authoring backgroundSource 🐞
- renable `autoHeight`

# TODO:
- Fix unclickable buttons.

[#179390297]
https://www.pivotaltracker.com/story/show/179390297

### DIFF
--- a/src/labbook/components/app.tsx
+++ b/src/labbook/components/app.tsx
@@ -26,7 +26,7 @@ const baseAuthoringProps = deepmerge(drawingToolBaseAuthoringProps, {
       backgroundSource: {
         title: "Background source",
         type: "string",
-        default: "any"
+        default: "snapshot"
       }
     }
   } as JSONSchema6,
@@ -46,9 +46,6 @@ const baseAuthoringProps = deepmerge(drawingToolBaseAuthoringProps, {
     },
     questionType: {
       "ui:widget": "hidden"
-    },
-    backgroundSource: {
-      "ui:widget": "hidden"
     }
   },
   // Just overwrite array, don't merge values.
@@ -66,7 +63,7 @@ export const App = () => (
   <BaseQuestionApp<IAuthoredState, IInteractiveState>
     Runtime={Runtime}
     baseAuthoringProps={baseAuthoringProps}
-    disableAutoHeight={true}
+    disableAutoHeight={false}
     isAnswered={isAnswered}
     linkedInteractiveProps={[{ label: "snapshotTarget", supportsSnapshots: true }]}
   />


dougmartin:
@knowuh have you tried to see what this looks like in the report mode?

tejal-shah:
@knowuh , I checked out your branch and ran this locally and while I see the authoring and the activity run (Uploading images and taking snapshots in a labbook activity locally) works well as a LARA Runtime item, when I run this as an AP activity, I get this error when I take a snapshot in the labbook :
`Snapshot has failed. Please try again.`
image

I am not sure if this error is because of the local env settings and might not be there when this is deployed to a proper deployed env. Have you see anything like this in your local setup?

scytacki:
@tejal-shah do image question interactive snapshots work in your local environment?
Based on the error I would initially guess they do not work either. 

However the snapshot service that is being used based on your screenshot is the same one used when I run this AP activity:
https://activity-player.concord.org/branch/master/?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Fapi%2Fv1%2Factivities%2F20814.json
And it seems to work fine.

So my next guess is that the snapshot service is actually returning an error message which doesn't include CORS headers, so then you get this misleading message instead of the actual error from the snapshot service.  If that is the case then this might be an error in how the labbook is working with the snapshot service. It could also be that the interactive you are trying to snapshot doesn't support snapshots.

This branch should be deployed so you could try it out with an activity on authoring.staging.concord.org that uses this branch of the question-interactives.

tejal-shah:
@scytacki , @knowuh , I added an image question that takes a snapshot from the tectonic explorer and it seems to be working okay with the activity player when using it in my local env of lara and local env of the question interactives. It seems to be a problem with only with the labbook qi. 
Also I am using the tectonic explorer as the interactive for snapshotting with both questions. They are both on separate pages with separate tectonic explorer interactives.

knowuh:
@tejal-shah Thank you for the report.  

I didn't run into those CORS issues when running from my local QI checkout, or when running from the branch deploy on authoring staging last week.

As per Scott's instructions I created a branch deploy version of the library as "Labbook branch fix-labbook-authoring" on authoring.staging:  https://authoring.staging.concord.org/library_interactives/430/edit

I wonder if this is an intermittent failure. 

I haven't hit the CORS issue running this activity:

https://activity-player.concord.org/branch/master/?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Fapi%2Fv1%2Factivities%2F21117.json&page=page_309259&runKey=e27f92a7-fe27-40ab-ba9e-87f0c3de9ea0

I need to check on the teacher report view to see how things look.  Maybe doing that will shed some more light on things.

knowuh:
@dougmartin Report view looks OK:

Screen Shot 2021-09-13 at 3 03 10 PM

knowuh:
Here is how it looks when launched from learn.staging as a student (using AP / master)

Screen Shot 2021-09-13 at 3 05 32 PM

tejal-shah:
@knowuh , Okay. I guess we should merge and deploy so we can test it in a staged env. Approving now.

tejal-shah (approved):

PR #135, opened 2021-09-29
Updated graph UI

sfentress:
https://www.pivotaltracker.com/story/show/179389391

Demo here: https://activity-player.concord.org/branch/master/?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Fapi%2Fv1%2Factivities%2F21318.json&page=page_310567

Authoring: https://authoring.staging.concord.org/activities/21318/pages/310567/edit

For the demo and authoring, I don't understand why the data often doesn't seem to connect.

### DIFF
--- a/src/graph/components/runtime.tsx
+++ b/src/graph/components/runtime.tsx
@@ -36,6 +36,10 @@ const getGraphOptions = (authoredState: IAuthoredState, { displayLegend }: { dis
     scales: {
       yAxes: [yAxis],
       xAxes: [{
+        scaleLabel: {
+          display: !!authoredState.xAxisLabel,
+          labelString: authoredState.xAxisLabel || "",
+        },
         ticks: {
           display: authoredState.displayXAxisLabels // this will show/hide labels only
         }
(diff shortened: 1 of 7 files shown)


src/graph/components/runtime.tsx
```diff
@@ -4,15 +4,40 @@ import {
...
+    },
+    ticks: {
+      beginAtZero: true,
+      maxTicksLimit: 5
+    }
+  };
+  if (authoredState.autoscaleYAxis === false) {
+    const maxY = typeof authoredState.yAxisMax === "number" ? authoredState.yAxisMax : 100;
+    // eslint-disable-next-line  @typescript-eslint/no-non-null-assertion
+    yAxis.ticks!.max = maxY;
```
pjanik, line 37: I think `!` and eslint-disable could be avoided by not providing `LinearScale` type in line 24 above.
`yAxis` would be type-checked anyway, lower down, when it's used in line 49. 
But I'm not 100% sure and the current approach is fine too if you don't have time to fiddle with it. 😉

pjanik (approved):
Looks good, thanks. 👍 One inline comment, but more fyi than necessary to change.

PR #136, opened 2021-09-30
Labbook: add thumbnail title to draw tool area

pjanik:
[#179583551]
Screenshot 2021-09-30 at 17 43 42

### DIFF
--- a/src/labbook/components/runtime.scss
+++ b/src/labbook/components/runtime.scss
@@ -60,6 +60,15 @@
   .draw-tool-wrapper {
     height: 500px;
     overflow: hidden;
+    position: relative;
+
+    .draw-tool-title {
+      top: 1px;
+      left: 42px;
+      border-color: var(--cc-charcoal-light-1);
+      border-top-left-radius: 0;
+      pointer-events: none;
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


knowuh (approved):
Looks good, thanks

(I was wondering if `classNames` would handle `classNames(className, ..., ...)` by gracefully handling undefined, but I wasn't sure it would work)

pjanik:
> (I was wondering if classNames would handle classNames(className, ..., ...) by gracefully handling undefined, but I wasn't sure it would work)

No, it needs to be a string. I tried this approach, but it'd look like that:

```javascript
if (!className) {
  className = "";
}
let finalClassName = classNames([css["thumbnail-title"]], {[css.empty]: empty, [className]: !!className });
```
and I thought it isn't very readable / clear. ;)

kswenson:
> (I was wondering if `classNames` would handle `classNames(className, ..., ...)` by gracefully handling undefined, but I wasn't sure it would work)

Yes, `classNames` handles `null`/`undefined` gracefully.

pjanik:
> > (I was wondering if `classNames` would handle `classNames(className, ..., ...)` by gracefully handling undefined, but I wasn't sure it would work)
> 
> Yes, `classNames` handles `null`/`undefined` gracefully.

Right, but as a value. Not as the property name - that's what I meant. In this particular case, the className variable acts as both.

PR #139, opened 2021-10-20
Side by Side interactive to support graph and DnD

sfentress:
This adds a new Side by Side interactive, which is much like the carousel except

1. It places two interactives side-by-side
2. If one of the interactives produces data and the other consumes it, it can pass the data between them

Demo in AP:

https://activity-player.concord.org/branch/master/?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Fapi%2Fv1%2Factivities%2F21318.json&page=page_310567

LARA editing:

https://authoring.staging.concord.org/activities/21318/pages/310567/edit

The data passing was done trying to affect the rest of the architecture as little as possible. One place where it changes existing code is to add a new property to the library-interactives list, `localLinkedInteractiveProp`, which is necessary so that the iframe-authoring / iframe-runtime can populate the linkedInteractives list. It's possible there is a different way to do this.

### DIFF
--- a/src/drag-and-drop/components/container.tsx
+++ b/src/drag-and-drop/components/container.tsx
@@ -145,10 +145,25 @@ export const Container: React.FC<IProps> = ({ authoredState, interactiveState, s
     }
   }, [authoredState.initialState?.itemPositions, authoredState.initialState?.targetPositions, setInitialState, setInteractiveState]);
 
-  const handleItemDrop = useCallback((targetData: IDropZone, targetPosition: IPosition, draggableItem: IDraggableItemWrapper) => {
+  const handleItemDrop = useCallback((targetData: IDropZone, targetPosition: IPosition, draggableItem: IDraggableItemWrapper, newItemPosition: IPosition) => {
     const droppedItem = draggableItem.item;
     const targetId = targetData.id;
     const targetDroppedItem = {targetId, targetPosition, droppedItem};
+
+    // nudge item into target. Prioritize the top left corner if the item is too small.
+    const nudgedItemPosition = {...newItemPosition};
+    if (nudgedItemPosition.left + draggableItem.rect.width > targetPosition.left + targetData.targetWidth) {
+      nudgedItemPosition.left = (targetPosition.left + targetData.targetWidth) - draggableItem.rect.width;
+    }
+    if (nudgedItemPosition.top + draggableItem.rect.height > targetPosition.top + targetData.targetHeight) {
+      nudgedItemPosition.top = (targetPosition.top + targetData.targetHeight) - draggableItem.rect.height;
+    }
+    if (nudgedItemPosition.left < targetPosition.left) {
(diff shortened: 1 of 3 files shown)


src/side-by-side/components/runtime.tsx
```diff
@@ -0,0 +1,83 @@
...
+
+  const handleNewInteractiveState = (interactiveStateSide: InteractiveStateSide, newInteractiveState: any) => {
+    setInteractiveState?.((prevState: IInteractiveState) => {
+      if (newInteractiveState.dataset && dataListeners.length > 0) {
+        dataListeners.forEach(listener => {
+          listener.phone.post("linkedInteractiveState", {
+            listenerId: listener.id,
+            interactiveState: {
+              dataset: newInteractiveState.dataset
+            }
```
pjanik, line 40:
```suggestion
            interactiveState: newInteractiveState
```
I think there's no reason to limit interactive state to dataset only.

src/side-by-side/components/iframe-authoring.tsx
```diff
@@ -0,0 +1,137 @@
...
+              name: "",
+              section: "interactive_box",
+              url: "",
+              thumbnailUrl: null,
+              supportsSnapshots: false
+            }
+          ]
+        }
+      );
+    });
```
pjanik, line 90: This means that interactive within side-by-side won't be able to link to other interactives. I guess we're not hoping to create something super generic, so I guess it's fine.

src/side-by-side/components/runtime.tsx
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  type InteractiveStateSide = "leftInteractiveState" | "rightInteractiveState";
+
+  const handleNewInteractiveState = (interactiveStateSide: InteractiveStateSide, newInteractiveState: any) => {
+    setInteractiveState?.((prevState: IInteractiveState) => {
+      if (newInteractiveState.dataset && dataListeners.length > 0) {
+        dataListeners.forEach(listener => {
+          listener.phone.post("linkedInteractiveState", {
+            listenerId: listener.id,
```
pjanik, line 37: It seems like it posts to all the interactives. But I guess somewhere along the way it's filtered by `listenerId`.

src/side-by-side/components/runtime.tsx
sfentress: Makes sense.

src/side-by-side/components/iframe-authoring.tsx
sfentress: Right, but this doesn't remove any existing feature. Currently they already can't link to other interactives -- e.g. a graph in a carousel could not link to an external interactive, because this IframePhone message was simply being dropped in the IframeRuntime.  This adds the new capability of being able to listen to the sibling interactive. If we want to add a further feature down the line of being able to listen to an external interactive, we could add that later.

PR #143, opened 2021-12-18
Updates slate-editor version to 0.7.3 to include the fix re multiple …

eireland:
…image dialogs when multiple slate editors exist.


kswenson (approved):
👍 LGTM

PR #146, opened 2022-01-10
Image interactive to show high-res image if regular image not available

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/178882818

[#178882818]

These changes make the image interactive use the high-res URL value (if present) if no URL is specified in the non high-res image field.

### DIFF
--- a/src/image/components/runtime.tsx
+++ b/src/image/components/runtime.tsx
@@ -22,6 +22,7 @@ export const Runtime: React.FC<IProps> = ({ authoredState }) => {
   const decorateOptions = useGlossaryDecoration();
   const { url, highResUrl, altText, caption, credit, creditLink, creditLinkDisplayText, scaling } = authoredState;
   const imageSize = useRef<IImageSize>();
+  const defaultImageUrl = url ? url : highResUrl;
 
   const getImageLayout = () => {
     switch (authoredState.scaling) {
@@ -46,7 +47,7 @@ export const Runtime: React.FC<IProps> = ({ authoredState }) => {
 
   const handleClick = () => {
     const allowUpscale = scaling === "fitWidth";
-    const modalImageUrl = highResUrl || url;
+    const modalImageUrl = highResUrl || defaultImageUrl;
     const title = `<strong>${caption || ""}</strong> <em>${credit || ""}</em> ${ReactDOMServer.renderToString(getCreditLink(false) || <span/>)}`;
     modalImageUrl && showModal({ type: "lightbox", url: modalImageUrl, isImage: true, allowUpscale, title });
     log("image zoomed in", { url });
@@ -72,7 +73,7 @@ export const Runtime: React.FC<IProps> = ({ authoredState }) => {
     <div className={css.runtime}>
       <div className={`${css.imageContainer} ${getImageLayout()}`} onClick={handleClick}>
         <img
-          src={url}
+          src={defaultImageUrl}
           alt={altText}
           title={altText}
           onLoad={getOriginalImageSize}
--- a/src/image/components/runtime.tsx
+++ b/src/image/components/runtime.tsx
@@ -1,4 +1,4 @@
-import React, { useRef } from "react";
+import React, { SyntheticEvent, useRef } from "react";
 import { IAuthoredState } from "./types";
 import { setSupportedFeatures, showModal } from "@concord-consortium/lara-interactive-api";
 import ReactDOMServer from "react-dom/server";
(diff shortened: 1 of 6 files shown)


src/image/components/runtime.tsx
```diff
@@ -46,7 +47,7 @@ export const Runtime: React.FC<IProps> = ({ authoredState }) => {
 
   const handleClick = () => {
     const allowUpscale = scaling === "fitWidth";
-    const modalImageUrl = highResUrl || url;
+    const modalImageUrl = highResUrl || defaultImageUrl;
```
kswenson, line 50: This change seems unnecessary -- it's essentially the same as `highResUrl || url || highResUrl`.

src/image/components/runtime.tsx
```diff
@@ -72,7 +73,7 @@ export const Runtime: React.FC<IProps> = ({ authoredState }) => {
     <div className={css.runtime}>
       <div className={`${css.imageContainer} ${getImageLayout()}`} onClick={handleClick}>
         <img
-          src={url}
+          src={defaultImageUrl}
```
kswenson, line 76:
```suggestion
          src={url || highResUrl}
```
The `defaultImageUrl` intermediate value doesn't seem to provide much value.

emcelroy: @kswenson Thanks -- you're right on both counts. I'm going to be taking a different approach though in light of a new understanding about what the PT story is asking for.

emcelroy:
@kswenson I'm wondering what you think of the latest changes. 

The code needs to use the high-res image URL not only as a fallback when the image URL is blank, but also if the image URL is an invalid URL. I added an error handler to accomplish that. It seems to work fine in the browser. The automated test I added for it fails, though. I don't know why but the test code image's `src` value doesn't change when the error handler is triggered.

I'm sure the error handler is being triggered when the test is run. If I add console log statements to the handler function, the log messages appear during the test. But the test never picks up the change to the image's `src` value. It always has the original value.

I've tried various ways to make the test work, but have not had any luck. I also tried to simply test that the error handler function is called, but couldn't make that work either. I'm tempted to just omit the test altogether, but was hoping to get your thoughts before I do that. Does anything seem obviously wrong to you with the test or maybe the error handler?

kswenson:
> @kswenson I'm wondering what you think of the latest changes.
> 
> The code needs to use the high-res image URL not only as a fallback when the image URL is blank, but also if the image URL is an invalid URL. I added an error handler to accomplish that. It seems to work fine in the browser. The automated test I added for it fails, though. I don't know why but the test code image's `src` value doesn't change when the error handler is triggered.
> 
> I'm sure the error handler is being triggered when the test is run. If I add console log statements to the handler function, the log messages appear during the test. But the test never picks up the change to the image's `src` value. It always has the original value.
> 
> I've tried various ways to make the test work, but have not had any luck. I also tried to simply test that the error handler function is called, but couldn't make that work either. I'm tempted to just omit the test altogether, but was hoping to get your thoughts before I do that. Does anything seem obviously wrong to you with the test or maybe the error handler?

The handler code looks fine; I suspect the problem is with the test. The `onError` callback is probably asynchronous, so I bet what's happening is that your `expect` is being called before the `src` change has taken effect. The solution is presumably one of the usual approaches to unit-testing async code. The other possibility that occurs to me is that it may require a re-render for the effect to be seen, but that seems less likely than the simple async test explanation.

emcelroy:
@kswenson Thanks for the suggestion. You're right that it's an async issue. In light of that, I spent more time trying to get the test to work, but still had a lot of trouble. The only thing that would work was to wrap the `expect` in a `setTimeout` which I suspect is less than ideal even though it works. I've spent more time on this than I should have, though, and think I should call it quits with that. Would you say it's at least acceptable?

emcelroy:
@scytaki I wasn't able to add test coverage for the code I added to the image interactive. @kswenson helped me get further along with that than I could on my own, but it still wasn't clear how to make a working test for the image error handler. Kirk also discovered that two other tests weren't working correctly. He fixed one and disabled the other.

I've spent a good chunk of time on this and think I should leave it as is for now. The new feature appears to work fine in the browser. The high-res image is shown if the other image URL isn't a valid URL.

Is how we've left things acceptable? Is there anything else we should do to ensure the test coverage is followed up on at some point?

scytacki (approved):
Seems good to me.  Too bad about the tests, but I understand, and I agree it isn't worth spending more time on it.

PR #147, opened 2022-01-12
Open Response default text should stay

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180721068

[#180721068]

These changes put `defaultAnswer` content (when present) into the textarea's `value` property instead of its `placeholder` property. Default answer content is meant to remain as part of the student's response instead of being a temporary placeholder.


pjanik (approved):
Looks good 👍

PR #149, opened 2022-01-13
Fix TE correct & distractor markers don't line up with radio buttons

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/180414126

[#180414126]

This fixes a bug viewable only in Firefox where Teacher Edition correct and distractor markers for multiple choice questions don't line up correctly with their associated radio buttons.

(The missing question number heading in the local AP screen grab below is an issue unrelated to the change made in this PR.)

image


scytacki (approved):
looks good to me

PR #150, opened 2022-01-18
Opens prompt links in a new tab

eireland:
demo [here](https://activity-player.concord.org/branch/master/?activity=https%3A%2F%2Fauthoring.staging.concord.org%2Fapi%2Fv1%2Factivities%2F21419.json&page=page_311012) Question 10. Note that Question 8 still does not work since this is still pointing to QI master branch.

### DIFF
--- a/src/shared/utilities/render-html.ts
+++ b/src/shared/utilities/render-html.ts
@@ -15,6 +15,14 @@ interface DomElement {
 // eslint-disable-next-line @typescript-eslint/ban-types
 export type ParseHTMLReplacer = (domNode: DomElement) => JSX.Element | object | void | undefined | null | false;
 
+DOMPurify.setConfig({ ADD_ATTR: ['target'] });
 export function renderHTML(html: string, replace?: ParseHTMLReplacer) {
-  return parse(DOMPurify.sanitize(html || ""), { replace });
+  let newHtml;
+  if (html.includes("<a")) {
+    const position = html.indexOf("<a") + 2;
+    newHtml = [html.slice(0, position), " target=\"_blank\"", html.slice(position)].join("");
+  } else {
+    newHtml = html;
+  }
+  return parse(DOMPurify.sanitize(newHtml || ""), { replace });
 }
--- a/src/shared/utilities/render-html.ts
+++ b/src/shared/utilities/render-html.ts
@@ -16,13 +16,15 @@ interface DomElement {
 export type ParseHTMLReplacer = (domNode: DomElement) => JSX.Element | object | void | undefined | null | false;
 
 DOMPurify.setConfig({ ADD_ATTR: ['target'] });
-export function renderHTML(html: string, replace?: ParseHTMLReplacer) {
-  let newHtml;
-  if (html.includes("<a")) {
-    const position = html.indexOf("<a") + 2;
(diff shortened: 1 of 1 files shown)


emcelroy (commented):
I was going to say we should also add `rel="noopener` to the link since authors will likely link to websites we don't control (see https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener), but then saw  https://mathiasbynens.github.io/rel-noopener/ has been updated to say "browsers now implicitly set rel=noopener for any target=_blank link." 

According to https://caniuse.com/mdn-html_elements_a_implicit_noopener that's true of most  browsers now, but there are some laggards. IE will probably never be updated. Not sure we have to worry about that or Opera or Android Browser...

I guess I would add `rel="noopener` just to be safe. It's probably not critical though if it's difficult for some reason.

Otherwise this looks fine and works as expected.

src/shared/utilities/render-html.ts
```diff
@@ -15,6 +15,14 @@ interface DomElement {
...
+DOMPurify.setConfig({ ADD_ATTR: ['target'] });
 export function renderHTML(html: string, replace?: ParseHTMLReplacer) {
-  return parse(DOMPurify.sanitize(html || ""), { replace });
+  let newHtml;
+  if (html.includes("<a")) {
+    const position = html.indexOf("<a") + 2;
+    newHtml = [html.slice(0, position), " target=\"_blank\"", html.slice(position)].join("");
+  } else {
+    newHtml = html;
+  }
```
kswenson, line 26: This kind of string manipulation is brittle and error-prone. It would be better to enforce this at the node level rather than via string manipulation. There is probably a way to accomplish this via the `replace` option of the parser but in this case I think hooking into `DOMPurify` via the `afterSanitizeAttributes` hook may make more sense. I came across an example that looked like this:
```javascript
DOMPurify.addHook("afterSanitizeAttributes", function(node) {
  // Fix elements with `target` attribute:
  // - allow only `target="_blank"
  // - add `rel="noopener noreferrer"` to prevent https://www.owasp.org/index.php/Reverse_Tabnabbing

  const target = node.getAttribute("target");
  if (isString(target) && target.toLowerCase() === "_blank") {
    node.setAttribute("rel", "noopener noreferrer");
  } else {
    node.removeAttribute("target");
  }
});
```
In this case I think it would look something more like (note that this is untested code, so you'll presumably have to futz with it to get it right):
```javascript
DOMPurify.addHook("afterSanitizeAttributes", function(node) {
  // Fix `a` elements:
  // - add `target="_blank"
  // - add `rel="noopener noreferrer"` to prevent https://www.owasp.org/index.php/Reverse_Tabnabbing
  if (node.tagName.toLowercase() === "a") {
    node.setAttribute("target", "_blank");
    node.setAttribute("rel", "noopener noreferrer");
  }
});
```
Separately, I've thought about publishing `renderHtml()` as a reusable npm module, in which case this kind of configuration should be client-configurable rather than being built into the function, but that possibility can be left for another day.

kswenson (changes requested):
I think there's a better approach that I detailed below.

kswenson (approved):
👍 Changes look good (assuming they work 😉 )

PR #152, opened 2022-01-19
Adds caption text field to videos

eireland:
Demo [here](https://authoring.staging.concord.org/activities/21419/pages/311023)


emcelroy (changes requested):
This looks good except that the caption and credit shouldn't be centered. To keep things consistent, they should be left justified as in the original video assessment item and the Library Interactive Image Interactive. (I see that the credit was already being centered before you made changes, but I'm pretty sure that was a mistake.)

emcelroy (approved):
Looks good, thank you!

PR #153, opened 2022-01-27
Changes video player name in conversion script to match template doc

eireland:
There's was a problem with how the question interactives conversion script was converting videos. Template document was pointing the the wrong interactive during conversion, and names didn't match type map and so script was not seeing the template. This fix makes the conversion script match the template activity.


emcelroy (changes requested):
For such a small difference, I think changing the name of the library interactive in LARA is a better option than changing the conversion script code. It'll take much less time.

I went ahead and changed the library interactive name so it should work with "Video Player" now. Sorry I didn't think to do that earlier when I fixed the template activity.

emcelroy (approved):
Looks good -- thanks!

PR #156, opened 2022-02-03
 Support interactiveApi with the full-screen Scaling interactive [#181103123]

dougmartin:
This change upgrades the lara client client to the current version and passes the full initInteractive message passed by AP to the fullscreen interactive mixed in with the current initMessage the fullscreen interactive creates.  The interactiveApi in CFM needs the info in the full message.

It was decided in this PR to only address the fullscreen interactive.  This same init message passing may need to be done for the other interactives that use the iframe-runtime container.

### DIFF
--- a/src/full-screen/components/runtime.tsx
+++ b/src/full-screen/components/runtime.tsx
@@ -93,7 +93,7 @@ export const Runtime: React.FC = () => {
 
   if (!subinteractiveUrl) {
     return <div>No sub items available. Please add them using the authoring interface.</div>;
-  } else {
+  } else if (initMessage) {
     return (
       <>
         <IframeRuntime url={subinteractiveUrl}
@@ -101,10 +101,13 @@ export const Runtime: React.FC = () => {
                        interactiveState={interactiveState}
                        setInteractiveState={setInteractiveState}
                        setHint={setHint}
+                       initMessage={initMessage}
         />
         {screenfull &&
           <FullScreenButton isFullScreen={screenfull.isFullscreen} handleToggleFullScreen={toggleFullScreen} />}
       </>
     );
+  } else {
+    return <div>Loading...</div>;
   }
 };
--- a/src/full-screen/components/runtime.tsx
+++ b/src/full-screen/components/runtime.tsx
@@ -90,8 +90,7 @@ export const Runtime: React.FC = () => {
   if (!wrappedInteractive) {
     return <div>No sub items available. Please add them using the authoring interface.</div>;
   } else {
-    const shared = queryString.parse(location.hash)?.shared;
-    const url = `${wrappedInteractive}${shared ? `#shared=${shared}` : ""}`;
+    const url = Array.isArray(wrappedInteractive) ? wrappedInteractive[0] : wrappedInteractive;
 
     return (
       <>
--- a/src/full-screen/components/runtime.tsx
+++ b/src/full-screen/components/runtime.tsx
@@ -85,29 +85,25 @@ export const Runtime: React.FC = () => {
 
   const iframeStyle = setScaling();
 
-  const params = queryString.parse(location.search);
-  const hash = queryString.parse(location.hash);
-  const sharedDoc = hash.shared ? "#shared="+hash.shared : "";
-  const preUrl = params?.wrappedInteractive + sharedDoc;
-  const subinteractiveUrl = preUrl && encodeURI(preUrl);
+  const wrappedInteractive = queryString.parse(location.search)?.wrappedInteractive;
 
-  if (!subinteractiveUrl) {
+  if (!wrappedInteractive) {
     return <div>No sub items available. Please add them using the authoring interface.</div>;
-  } else if (initMessage) {
+  } else {
+    const shared = queryString.parse(location.hash)?.shared;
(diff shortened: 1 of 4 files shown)


src/shared/components/iframe-runtime.tsx
```diff
@@ -91,7 +92,9 @@ export const IframeRuntime: React.FC<IProps> =
       if (libraryInteractive?.localLinkedInteractiveProp && authoredState?.[libraryInteractive.localLinkedInteractiveProp]) {
         linkedInteractives = [ {id: authoredState[libraryInteractive.localLinkedInteractiveProp], label: libraryInteractive.localLinkedInteractiveProp} ];
       }
+
       phone.post("initInteractive", {
+        ...initMessage,   // for now only fullscreen sets this prop so that the needed info for cfm interactiveApi is passed
```
kswenson, line 97: When I was looking at this with @eireland, I/we had a question about the contents of the initMessage. The property at issue is that the `interactiveApiProvider` assumes that there is an `interactive` property which can be dereferenced for its `id`, presumably for the purpose of getting the `id` of the embedded interactive. When this is sent from the AP in a situation in which there is a full-screen interactive wrapping another interactive, which interactive is specified in the `interactive` prop? Is it the full-screen interactive or the embedded one? If it's the former, then shouldn't the full-screen interactive be replacing the `interactive` property with one that references the embedded interactive instead?

scytacki: In the case of the full-screen interactive, it is acting basically as a transparent proxy of the interactive it is wrapping. So I think it makes the most sense for the interactive it is wrapping to be passed the id that was originally passed by the AP (host). 

Furthermore, the host doesn't know anything about the wrapped interactive so it has no id that it could provide. This id basically represents the id of the interactive in the authoring system. 

For some tangential info about these IDs they are mentioned in a section of this doc: https://docs.google.com/document/d/1vhRfcrxPRMtJRkHDqYbDPXKPDCqZV5JjjKoSmqB5zMg/edit#heading=h.3ig3ztmzycvi
Note however, that since that section was written, the format of the ids has been updated as indicated in a following section: https://docs.google.com/document/d/1vhRfcrxPRMtJRkHDqYbDPXKPDCqZV5JjjKoSmqB5zMg/edit#heading=h.m617becjaecd

Coming back to the original problem, when the interactive being wrapped is being used in the carousel or side-by-side  which support wrapping multiple interactives the id is a problem. With Doug's change the id is still not set when this is used by the carousel or side-by-side.  It isn't clear what to should be set to in these cases. It seems like it should be unique for each of the interactives. However, the current way the logging and reporting services work it is probably best that it is the same as the top level interactive. That is because these systems will only know about this top level id, so when they are trying to group data together they will only find the data if it has this top level id.
Another potential issue is how attachments work. I believe the attachment code on the host and in the interactive depends on this id being unique. This way attachments saved by one interactive do not conflict with attachments with the same name saved by another interactive. 

So the summary is that this id issue is problematic for these nested interactives. Hopefully we'll run into this when working on the reporting of the side-by-side interactive which is scheduled next week. And we can figure out how bad the problem is then.

scytacki (approved):
Hopefully this will fix the problem for now. 
We will very likely need to come back to it soon.

dougmartin:
@scytacki I added a couple of commits since your approval:

- https://github.com/concord-consortium/question-interactives/pull/156/commits/3788936d7059217ae3832210cd6e9623ef44f2db : this removes the escaping of the url in the iframe runtime
- https://github.com/concord-consortium/question-interactives/pull/156/commits/4b31f0ff881f5986018ece63be1454458bf81ac4 : this fixes a line to set the `initMessage` that got dropped in the previous commit

src/full-screen/components/runtime.tsx
```diff
@@ -85,18 +85,17 @@ export const Runtime: React.FC = () => {
...
-  const subinteractiveUrl = preUrl && encodeURI(preUrl);
+  const wrappedInteractive = queryString.parse(location.search)?.wrappedInteractive;
 
-  if (!subinteractiveUrl) {
+  if (!wrappedInteractive) {
     return <div>No sub items available. Please add them using the authoring interface.</div>;
-  } else if (initMessage) {
+  } else {
+    const shared = queryString.parse(location.hash)?.shared;
+    const url = `${wrappedInteractive}${shared ? `#shared=${shared}` : ""}`;
```
scytacki, line 94: I think this should be removed. I know it was there before, but that seems like a mistake to me. 
The author should be escaping the URL correctly so there would be no need for the interactive to look for hash parameters on the full-screen interactive's URL.

Ie. the URL should have a search string like `?wrappedInteractive=https%3A%2F%2Fcodap.concord.org%23shared%3D1234`

Which would then get turned into `https://codap.concord.org#shared=1234`
`queryString.parse`

This way there is no special logic in the full-screen interactive, it is just passing through the wrappedInteractive parameter. 

No one is using this full-screen interactive so it is fine to change how it works.

scytacki (changes requested):
I think it should be simplified even further.

src/full-screen/components/runtime.tsx
dougmartin: Resolved in https://github.com/concord-consortium/question-interactives/pull/156/commits/8bc09db1edfc1d38a7469952bfec980f817ffda1

PR #161, opened 2022-02-15
Fix missing environment variable in production builds [#181270128]

dougmartin:
The S3_DEPLOY_DIR environment variable was not set for production deploys causing reads of s3_website.yml to fail due to it trying to call Regexp.escape(nil) where nil was the missing environment variable value.

With this fix the S3_DEPLOY_DIR is always exported.


scytacki (approved):
Seems fine to me.

PR #162, opened 2022-02-16
Added version.json generation in webpack.

dougmartin:
This is so we can see which version is live in the root production deploy either manually or programmatically.

This generates a root level file at compile time containing the following information pulled either from the package.json file or the current date:

- name
- version
- description
- buildDate (human readable)
- buildTimestamp (unix timestamp)

**UPDATE**

Here is the generated version.json for this branch:

https://models-resources.concord.org/question-interactives/branch/generate-version-json/version.json

### DIFF
--- a/generate-version-json.js
+++ b/generate-version-json.js
@@ -1,0 +1,13 @@
+const {name, version , description} = require('./package.json');
+
+const now = new Date();
+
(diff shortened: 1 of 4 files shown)


scytacki (approved):
looks fine to me

PR #163, opened 2022-02-16
Fix drag-and-drop interactive does not look correct in report/dashboard.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181154848

[#181154848]

When the drag-and-drop interactive is being displayed within a teacher dashboard or report, it will be scaled to fit the confines of its container and not display its prompt. If the optional `view` URL param is present and set to "standalone", the interactive should not be scaled and its prompt should be displayed.

These changes go along with [these changes to portal-report](https://github.com/concord-consortium/portal-report/pull/377) and [these changes in LARA](https://github.com/concord-consortium/lara/pull/909).

### DIFF
--- a/src/drag-and-drop/components/container.tsx
+++ b/src/drag-and-drop/components/container.tsx
@@ -13,7 +13,7 @@ import { cssUrlValue } from "../../shared/utilities/css-url-value";
 export interface IProps extends IRuntimeQuestionComponentProps<IAuthoredState, IInteractiveState> {
   // Used only for authoring (initial state is part of the authored state).
   setInitialState?: (initialState: IInitialState) => void;
-  view?: string;
+  view?: "standalone";
 }
 
 interface IDimensions {
--- a/src/drag-and-drop/components/container.tsx
+++ b/src/drag-and-drop/components/container.tsx
@@ -13,6 +13,7 @@ import { cssUrlValue } from "../../shared/utilities/css-url-value";
 export interface IProps extends IRuntimeQuestionComponentProps<IAuthoredState, IInteractiveState> {
   // Used only for authoring (initial state is part of the authored state).
   setInitialState?: (initialState: IInitialState) => void;
+  view?: string;
 }
 
 interface IDimensions {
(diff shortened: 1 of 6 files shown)


src/drag-and-drop/components/container.tsx
```diff
@@ -226,7 +230,9 @@ export const Container: React.FC<IProps> = ({ authoredState, interactiveState, s
   const draggingAreaStyle = {
     width: canvasWidth + "px",
     height: canvasHeight + "px",
-    backgroundImage: authoredState.backgroundImageUrl ? cssUrlValue(authoredState.backgroundImageUrl) : undefined
+    backgroundImage: authoredState.backgroundImageUrl ? cssUrlValue(authoredState.backgroundImageUrl) : undefined,
+    transform: report && view !== "standalone" ? `scale(${scaleRatioForReport})` : undefined,
```
scytacki, line 234: I'm guessing you tested this, but often when applying scale changes in css mouse pointer and touch events can get messed up. You'd usually see the issue when trying to click and drag something. 
This issue is why we usually do scaling at the iframe level instead of inside of the interactive.  But in this case this approach seems to make sense, as long as we test it and fix any pointer event issues.

src/shared/components/base-question-app.tsx
```diff
@@ -30,6 +30,7 @@ export interface IRuntimeQuestionComponentProps<IAuthoredState, IInteractiveStat
   interactiveState?: IInteractiveState | null,
   setInteractiveState?: (updateFunc: UpdateFunc<IInteractiveState>) => void;
   report?: boolean;
+  view?: string;
```
scytacki, line 33: This should be restricted to "standalone".  It is probably best to just explicitly put that here for now.

But if you wanted to delve into typescript stuff, you can probably find a way to pull the possible values out of the lara-interactive-client types.

scytacki (changes requested):
Should add explicit "standalone" value here.

src/drag-and-drop/components/container.tsx
emcelroy: @scytacki In the context of a teacher report/dashboard, do we want to allow the interactive to be manipulated? When working on this it seemed like the interactive was specifically frozen to not allow it to be changed, but maybe I'm wrong about that.

scytacki (changes requested):
The code looks good to me.

However the dependency on the lara-interactive-api needs to be update to reference a new release. Without this the Jest tests won't pass on CI.

So the lara PR should be merged and a new release of the lara-interactive-api should be released to npm. Then this PR should be updated to refer to that new version.

scytacki (approved):
Looks good.

I would have bumped the api version 1.6.0 because this was a new feature of the API not a bug fix, but it is fine it isn't worth re-releasing.

PR #164, opened 2022-02-18
Fix side-by-side interactive doesn't fully update state consistently

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181154948

[#181154948]

These changes are meant to make the side-by-side interactive call data listeners immediately after they're added and provide any state of its child interactives that it knows about. Without that feature, it's possible for the child interactives to initially load without any state.

### DIFF
--- a/src/side-by-side/components/runtime.tsx
+++ b/src/side-by-side/components/runtime.tsx
@@ -29,10 +29,10 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
   }, [dataListeners]);
 
   useEffect(() => {
-    if (interactiveState?.leftInteractiveState.dataset) {
+    if (interactiveState?.leftInteractiveState?.dataset) {
       updateDataListeners(interactiveState.leftInteractiveState);
     }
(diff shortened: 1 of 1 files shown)


scytacki (changes requested):
With the minor consistency change this seems OK to me.

It still doesn't handle the general problem of an interactive in the side-by-side interactive wanting to listen to state changes in any other interactive. It is just working specifically in the the case where one of the interactives has a dataset property in its state. 
I put a little info about that in this document where I'm trying to keep track of all of the interactive hosts and how they vary: https://docs.google.com/document/d/1TQHCQ0ULontviH-wv7r-3I0n8HEdXOb5U_8AKtzQoT4/edit#

The general solution would be more complicated, so I think it is fine to keep this as is for now.

pjanik (approved):
It's been a while since I was working on that, but I skimmed through the code and it looks good. I'd ask if we need to check to `dataset` prop, but I bet it would be too easy to make all that generic (I don't remember too well the linked state implementation).

PR #167, opened 2022-03-01
Fix side-by-side interactive in report or dashboard layout issues.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181220233

[#181220233]

### DIFF
--- a/src/shared/components/iframe-runtime.tsx
+++ b/src/shared/components/iframe-runtime.tsx
@@ -25,13 +25,12 @@ interface IProps {
   setHint?: (state: any) => void | null;
   addLocalLinkedDataListener?: (request: IAddLinkedInteractiveStateListenerRequest, phone: IframePhone) => void;
   scale?: number;
-  view?: "standalone";
 }
 
 export const IframeRuntime: React.FC<IProps> =
   ({ authoredState, id, iframeStyling, interactiveState, logRequestData, report,
       url, setHint, setInteractiveState, addLocalLinkedDataListener, initMessage, 
-      scale, view }) => {
+      scale }) => {
     const [ iframeHeight, setIframeHeight ] = useState(300);
     const [ internalHint, setInternalHint ] = useState("");
     const iframeRef = useRef<HTMLIFrameElement>(null);
@@ -99,7 +98,6 @@ export const IframeRuntime: React.FC<IProps> =
       const initInteractiveMessage = {
         ...initMessage,   // for now only fullscreen sets this prop so that the needed info for cfm interactiveApi is passed
         mode: report ? "report" : "runtime",
-        view: view,
         authoredState,
         // This is a trick not to depend on interactiveState.
         interactiveState: interactiveStateRef.current,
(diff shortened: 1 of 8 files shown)


emcelroy:
@scytacki I haven't tested this with a wide variety of side-by-side interactive examples yet. I wanted to get your thoughts on the general approach first.

It will require adding yet another property (`division`) to `IReportInitInteractive` in the lara-interactive-api. I tried a few ways of scaling things at a higher level so we wouldn't have to do that, but I couldn't get anything else to work correctly.

Also note especially that I added a `renderRuntime` method to the `BaseApp` component. Originally, the `BaseApp` component had a comment about it being for interactives that don't appear in reports. But since the graph interactive uses BaseApp and will appear in reports, that isn't true. Would that change be OK? Or would it be better to modify the graph interactive to use `BaseQuestionApp`?

scytacki (changes requested):
I think we have to find another way instead of using a new division property. That is too specific to the side-by-side interactive.  I can work with you on this later today. Have you tried using the dimensions of the window to figure out the width of the iframe?
Or possibly having the side-by-side apply scaling to the iframe element instead of requiring the graph and drag-drop to do their own scaling.

If we can get rid of that division property maybe that means we can also avoid having to add the report stuff to the base-app. But if we have to, then I guess it is OK. We should just add a more specific comment about why BaseApp supports reporting. Since generally if an interactive doesn't save state it can't show up in the report.

src/shared/components/iframe-runtime.tsx
```diff
@@ -117,9 +121,24 @@ export const IframeRuntime: React.FC<IProps> =
         phoneRef.current.disconnect();
       }
     };
-  },[addLocalLinkedDataListener, authoredState, logRequestData, report, setHint, url, initMessage]);
+  },[addLocalLinkedDataListener, authoredState, logRequestData, report, setHint, url, initMessage, view]);
 
-  const iframeStyle = iframeStyling ?? {width: "100%", height: iframeHeight, border: "none"};
+  let scaledIframeStyle = undefined;
+  if (scale && report && view !== "standalone") {
```
scytacki, line 127: Rather than pass the `view` in here. It would be better for the component using this IframeRuntime to not set the scale when the view is standalone. That way this IframeRuntime component doesn't need to know about this view property.

src/shared/components/iframe-runtime.tsx
```diff
@@ -117,9 +121,24 @@ export const IframeRuntime: React.FC<IProps> =
...
+      transformOrigin: "left top",
+      width: `${100/scale}%`
+    };
+  }
+
+  const iframeStyle = scaledIframeStyle
+                        ? scaledIframeStyle
+                        : iframeStyling
+                          ? iframeStyling
+                          : {width: "100%", height: iframeHeight, border: "none"};
```
scytacki, line 141: I would expect that iframeStyling should override the scale setting. So if a user of the IframeRuntime component sets the iframeStyling this would override the scale based styling.

src/shared/components/iframe-runtime.tsx
```diff
@@ -117,9 +121,24 @@ export const IframeRuntime: React.FC<IProps> =
...
     };
-  },[addLocalLinkedDataListener, authoredState, logRequestData, report, setHint, url, initMessage]);
+  },[addLocalLinkedDataListener, authoredState, logRequestData, report, setHint, url, initMessage, view]);
 
-  const iframeStyle = iframeStyling ?? {width: "100%", height: iframeHeight, border: "none"};
+  let scaledIframeStyle = undefined;
+  if (scale && report && view !== "standalone") {
+    scaledIframeStyle = {
+      border: "none",
+      height: `${300/scale}px`,
```
scytacki, line 130: It would be better if `iframeHeight` was used here instead of 300. This way interactives that use the interactive api to request more space will still work properly.

scytacki (changes requested):
looks good. I just have a few requests to keep the iframeruntime general.

src/shared/components/iframe-runtime.tsx
emcelroy: @scytacki Do the latest changes make sense? I had thought of doing that before, but lines 89 and 103 in runtime.tsx seemed a little on the too cognitively complex side. Maybe not, though? Or maybe there's a better way to do it that I'm not thinking of?

scytacki: I think the approach in runtime.tsx is OK.
You could pull it out to a `leftScale` and `rightScale` variable so the logic is not inside of the jsx. For example:
```
let leftScale = undefined;
let rightScale = undefined;
if (report && view !== "standalone") {
  leftScale = leftReportScale;
  rightScale = rightReportScale;
}
```
But I think what you've got might be easier to read than what I put above.

scytacki (approved):
Looks great to me

PR #168, opened 2022-03-02
Fix question text cut off when image appears in question prompt

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181427575

[#181427575]

Adds style rules to ensure images in an interactive's prompt text will not be wider than their container. This will prevent large images from causing interactive content to be cut off by a narrow containing column. 

It's possible the `height: auto` rule will prevent authors from setting an image to a specific height, but I think it's unlikely authors will need to do that. And it's probably more important to maintain an image's aspect ratio so it doesn't get distorted.

image


kswenson (approved):
👍 Seems okay to me

PR #169, opened 2022-03-03
181436055 update dependencies

emcelroy:
Note that I upgraded Firebase to Firebase 9 but opted to use the [compat library instead of the new modular library](https://firebase.google.com/docs/web/modular-upgrade). I ran into trouble trying to use the modular library and decided it wasn't worth the time to convert things right now.


src/video-player/components/runtime.tsx
```diff
@@ -57,55 +57,57 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
...
+      });
 
-    return () => {
-      player.dispose();
-    };
+      return () => {
+        player.dispose();
+      };
+    }
   }, [authoredState]);
```
dougmartin, line 111: Need to add a dependency to `playerRef.current` in the array

dougmartin (changes requested):
Need to fix effect dependency

src/shared/hooks/use-linked-interactives-authoring.test.ts
```diff
@@ -108,7 +108,7 @@ describe("useLinkedInteractives", () => {
         schema: {}
       });
     };
-    renderHook(HookWrapper);
+    act(() => { renderHook(HookWrapper); });
```
kswenson, line 111: In my experience, `act()` is normally used for things you do to a component _after_ it's rendered, e.g.
```
  renderHook(...);
  act(() => userEvent.click());
```
Was there a particularly reason you needed to wrap the renders themselves in `act()`?

package.json
```diff
@@ -68,100 +68,104 @@
...
-    "babel-jest": "^26.3.0",
-    "babel-loader": "^8.1.0",
+    "babel-jest": "^26.6.3",
+    "babel-loader": "^8.2.3",
     "canvas": "2.6.1",
-    "css-loader": "^4.2.2",
-    "cypress": "^6.0.0",
+    "css-loader": "^4.3.0",
+    "cypress": "^9.5.1",
     "enzyme": "^3.11.0",
```
kswenson, line 97: You replaced `enzyme` with `@testing-library` in at least one instance below. Is `enzyme` still used elsewhere or could it be removed at this point?

src/video-player/components/runtime.tsx
```diff
@@ -57,55 +57,57 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
...
-              log("cue change", { videoUrl: authoredState.videoUrl, currentCue: (currentCue as VTTCue).text });
-            });
+    if (playerRef.current) {
+      const player: videojs.Player = videojs(playerRef.current,
+        {
+          controls: true,
+          fluid: !(authoredState.fixedAspectRatio || authoredState.fixedHeight),
+          // This is a new property not supported by the current types
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore
```
kswenson, line 67: This is not new code, so feel free to ignore this, but if you use `@ts-expect-error` here then it'll notify when the types are updated to support the new property.

src/video-player/components/runtime.tsx
kswenson: > Need to add a dependency to `playerRef.current` in the array

Maybe -- refs are generally used specifically to store values that shouldn't trigger re-renders on change. I take your point to be that the effect should run again when `playerRef.current` changes (which makes sense) but before reflexively adding `playerRef.current` to the dependencies I'd ask whether the value should really be in a ref. If it should trigger a re-render on change then maybe it should be state, for instance.

tsconfig.json
```diff
@@ -10,7 +10,7 @@
     "resolveJsonModule": true,
     "experimentalDecorators": true,
     "strictNullChecks": true,
-    "types": ["jest"],
+    "types": ["jest", "@testing-library/jest-dom"],
```
kswenson, line 13: The problem with this is that it makes these types available everywhere, but we only want them available in test code. It should be an error to reference these globals/types in non-test code. The conventional way to handle this for `@testing-library/jest-dom` is to import it in the `setupTests.ts` file.

webpack.config.js
```diff
@@ -223,6 +214,9 @@ module.exports = (env, argv) => {
       new GenerateJsonFromJsPlugin({
         path: './generate-version-json.js',
         filename: 'version.json'
+      }),
+      new ESLintPlugin({
+        extensions: ["tsx"]
```
kswenson, line 219: I would think you'd want to include ".ts" as well at the very least.

package.json
emcelroy: @kswenson  Enzyme is still required by a bunch of tests that use its `mount` and `shallow`. I guess we could change those to use React Testing Library instead, but I think that's probably beyond the scope of this work -- unless you think it's important?

kswenson: In the long run I think that should be the plan, but it's not a priority for this PR. I just saw that you changed it in one place and wanted to make sure that it hadn't become a stale dependency.

tsconfig.json
kswenson: I don't know exactly why it behaves the way it does, but I was able to fix it by:
1. Renaming `setupTests.js` to `setupTests.ts`
2. Making the same change in the jest configuration in `package.json`
3. Using `import `@testing-library/jest-dom` instead of `require`ing it

src/video-player/components/runtime.tsx
kswenson: I think what @dougmartin is pointing out is that in this code:
```typescript
  useEffect(() => {
    if (playerRef.current) {
      const player: videojs.Player = videojs(playerRef.current,
```
If `playerRef.current` were `undefined` at the time of the first render, then if it subsequently became defined we would want the effect to run again. The fact that things work currently suggests that either `playerRef.current` is always defined when the effect runs or that if it ever changes that change also correlates with other changes that trigger a re-render. I suspect that it's always defined on first render and the only time it becomes `undefined` is on unmount, in which case we don't really care whether the effect runs or not. The easiest thing to do in the short run is just to leave it alone since it seems to be working, but if we ever need to handle the case of transitioning from `undefined` to defined then we'd have to do something to make sure that the effect runs in that case.

tsconfig.json
emcelroy: @ksweson Thanks! I can confirm that fixes the problem.

src/video-player/components/runtime.tsx
emcelroy: OK, thanks. I've left the dependency array as is in the latest commit since it doesn't seem necessary to include `playerRef`. 

This may not need to be stated, but for the record, if I remove `if (playerRef.current) {`, the line immediately after it causes the error `Argument of type 'HTMLVideoElement | null' is not assignable to parameter of type 'string | Element'. Type 'null' is not assignable to type 'string | Element'.` to be thrown (where `Argument of type 'HTMLVideoElement | null'` refers to `playerRef.current`).

Let me know if you still think `playerRef` should be included, though, @dougmartin @lublagg.

kswenson:
> @dougmartin @lublagg FYI, my use of `act` here was unnecessary as @kswenson points out. I think I must've been trying to work out what exactly was triggering a warning that said a test should be using `act` and added it in too many places.

So I noticed while playing around that there were still a few places where there was an `act()`-related warning as well as another warning that seemed worth addressing. Let me know if you'd like me to just commit the changes, but in the interest of not introducing commit conflicts here are the differences:
```
diff --git a/src/image-question/components/app.test.tsx b/src/image-question/components/app.test.tsx
index 36891dc..92ed581 100644
--- a/src/image-question/components/app.test.tsx
+++ b/src/image-question/components/app.test.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { render } from "@testing-library/react";
+import { act, render } from "@testing-library/react";
 import { useAuthoredState, useInitMessage, useInteractiveState } from "@concord-consortium/lara-interactive-api";
 import { App } from "./app";
 import { IAuthoredState, IInteractiveState } from "./types";
@@ -50,7 +50,9 @@ describe("Image question", () => {
   it("renders in authoring mode", async () => {
     const { container } = render(<App />);
     expect(container).toBeDefined();
-    const promptEditor = await container.querySelector("#root_prompt");
-    expect(promptEditor?.className.includes("slate-editor")).toBe(true);
+    await act(async () => {
+      const promptEditor = await container.querySelector("#root_prompt");
+      expect(promptEditor?.className.includes("slate-editor")).toBe(true);
+    });
   });
 });
diff --git a/src/labbook/components/thumbnail-chooser/thumbnail-wrapper.test.tsx b/src/labbook/components/thumbnail-chooser/thumbnail-wrapper.test.tsx
index 4273e24..d3f253e 100644
--- a/src/labbook/components/thumbnail-chooser/thumbnail-wrapper.test.tsx
+++ b/src/labbook/components/thumbnail-chooser/thumbnail-wrapper.test.tsx
@@ -3,14 +3,12 @@ import { render, screen } from "@testing-library/react";
 import { IThumbnailProps } from "./thumbnail";
 import { IThumbnailWrapperProps, ThumbnailWrapper } from "./thumbnail-wrapper";
 
-const content: React.FC<IThumbnailProps> = () => <div data-testid="thumbnail">Thumbnail</div>;
-
 const thumb:IThumbnailProps = {
   id: "one",
   data: {},
   label: "one",
   empty: false,
-  thumbContent: content
+  thumbContent: "Thumbnail"
 };
 
 const emptyThumb:IThumbnailProps = {
@@ -18,7 +16,7 @@ const emptyThumb:IThumbnailProps = {
   data: {},
   label: "one",
   empty: true,
-  thumbContent: content
+  thumbContent: null
 };
 
 describe("ThumbnailWrapper component", () => {
diff --git a/src/open-response/components/app.test.tsx b/src/open-response/components/app.test.tsx
index 4711743..80da697 100644
--- a/src/open-response/components/app.test.tsx
+++ b/src/open-response/components/app.test.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import { render } from "@testing-library/react";
+import { act, render } from "@testing-library/react";
 import { useAuthoredState, useInitMessage, useInteractiveState
         } from "@concord-consortium/lara-interactive-api";
 import { App } from "./app";
@@ -51,7 +51,9 @@ describe("Open response question", () => {
   it("renders a rich text editor in authoring mode", async () => {
     const { container } = render(<App />);
     expect(container).toBeDefined();
-    const promptEditor = await container.querySelector("#root_prompt");
-    expect(promptEditor?.className.includes("slate-editor")).toBe(true);
+    await act(async () => {
+      const promptEditor = await container.querySelector("#root_prompt");
+      expect(promptEditor?.className.includes("slate-editor")).toBe(true);
+    });
   });
 });
```
The first two changes introduce `act()` where there's an asynchronous call happening from a `useEffect` so although the test code itself doesn't do anything that requires `act()` the asynchronous callback makes it appear that it does. The last change is related to the content being passed to the thumbnail which should just be the content rather than a component which renders the thumbnail.

emcelroy:
@kswenson Thanks -- I can incorporate those changes. 

I'm also all of a sudden seeing some Cypress test failures here that weren't showing up before and aren't showing up in my local environment now, either. I'll try to fix those as well.

src/shared/utilities/render-html.ts
```diff
@@ -12,8 +12,8 @@ interface DomElement {
     prev?: DomElement;
     type?: string;
 }
-// eslint-disable-next-line @typescript-eslint/ban-types
-export type ParseHTMLReplacer = (domNode: DomElement) => JSX.Element | object | void | undefined | null | false;
+
+export type ParseHTMLReplacer = (domNode: DomElement) => JSX.Element | Record<string, unknown> | void | undefined | null | false;
```
kswenson, line 16: 👍

kswenson (approved):
👍 LGTM!

PR #170, opened 2022-03-09
Added attachment support to full-screen wrapper [#181103123]

dougmartin:
This adds a proxy for the attachment messages between the wrapped interactive and the enclosing host (AP or Lara).

### DIFF
--- a/src/shared/components/iframe-runtime.tsx
+++ b/src/shared/components/iframe-runtime.tsx
@@ -93,6 +93,17 @@ export const IframeRuntime: React.FC<IProps> =
         // To ensure that this request isn't mixed up with an existing request the request id
         // is set to a random large id when it is passed to the enclosing host and then restored
         // when passed back to the client
+
+        //--------------------------------------
+        // FIXME: (from Scott during code review)
+        //
+        // A problem with this approach is the file names. This iframe-runtime is used by interactives that host multiple
+        // sub interactives like the side-by-side or carousel. Let's say there are two CFM sub interactives. They will each choose
+        // a file name of file.json so they will clobber each other since there will just be one folder managed by the parent interactive.
+        //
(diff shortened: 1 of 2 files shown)


src/shared/components/iframe-runtime.tsx
```diff
@@ -86,6 +86,23 @@ export const IframeRuntime: React.FC<IProps> =
           addLocalLinkedDataListener(request, phone);
         }
       });
+      phone.addListener("getAttachmentUrl", async (request: IAttachmentUrlRequest) => {
+        const client = getClient();
+
+        // This proxies the attachment request to the enclosing host (AP or Lara).
+        // To ensure that this request isn't mixed up with an existing request the request id
+        // is set to a random large id when it is passed to the enclosing host and then restored
+        // when passed back to the client
```
scytacki, line 95: A problem I just realized with this approach is the file names. This iframe-runtime is used by interactives that host multiple sub interactives like the side-by-side or carousel. Let's say there are two CFM sub interactives. They will each choose a file name of `file.json` so they will clobber each other since there will just be one folder managed by the parent interactive. 

This is a general problem broader then just this one case. Leslie is aware of it, so I think we'll take some time in the future to fix this up so we can properly support this multiple sub-interactives. 

But I think we should add a FIXME comment here indicating this is one place we are going to have to deal with.

scytacki (changes requested):
Looks good. I just requested a comment describing an issue that will happen multiple CFM interactives are used in a side-by-side or carousel.

src/shared/components/iframe-runtime.tsx
dougmartin: FIXME comment added in https://github.com/concord-consortium/question-interactives/pull/170/commits/7649fafd81d3265ec86a7ab79c3549cbda0fb515

scytacki (approved):
looks good

PR #176, opened 2022-04-04
Support multiple answer types and send them to the Portal Report

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181516279

[#181516279]

These changes add an open response report item.


src/open-response/components/report-item/init-message-info.tsx
```diff
@@ -0,0 +1,24 @@
...
+  // const {users, interactiveItemId, view} = initMessage.current;
+  // const userIds = Object.keys(users);
+  // const numAnswersAtInit = userIds.reduce<number>((acc, studentId) => {
+  //   return users[studentId].hasAnswer ? acc + 1 : acc;
+  // }, 0);
+  // const userAnswers = useRef<Record<string, any>>({});
+  // const numCurrentAnswers = Object.keys(userAnswers).length;
+
+  return (
+    // TODO: Figure out what, if anything, has to be shown here.
```
emcelroy, line 20: I modeled the new open response report item on the example interactive for attachments. Although the open response report item seems to be necessary for making the attachment information accessible on the portal-report side, it doesn't seem like it needs to render anything to be shown in the portal-report. Does that make sense, though?

dougmartin: Yes - I think this whole file can be removed along with the two summary componenents.

src/open-response/components/report-item/report-item.tsx
```diff
@@ -0,0 +1,58 @@
...
+          "Open Response Report Item Interactive: Unsupported version in getReportItemAnswer request:",
+          version
+        );
+      }
+    });
+
+    getClient().post("reportItemClientReady");
+  });
+
+  return (
```
dougmartin, line 50: I would just return null here and remove these two components - we don't need to render anything.

dougmartin (approved):
Looks good :+1:

scytacki (approved):
This seems good to me.

emcelroy:
@dougmartin @scytacki I've made the discussed changes and added some tests. I don't see a straightforward way to add more coverage for runtime.tsx or report-item.tsx. Is this OK as is? Or should I spend more time on adding coverage?

scytacki (approved):
Looks good to me.

I can see a way to add more coverage, but I don't think you need to do it. 
If you want to, basically you'd need to fill out the `addGetReportItemAnswerListener: jest.fn()` with more of an implementation that calls the passed in listener with different `request`s.

PR #177, opened 2022-04-04
Record OR - Log runtime events

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181223935

[#181223935]

These changes add logging for the recording of audio responses. There is also a small improvement to the handling of failed audio recordings which includes basic error notification in UI.

EDIT: I spent a little time trying to improve coverage for runtime.tsx so the recording and recording failure functionality were tested, but wasn't able to make that work.


src/open-response/components/runtime.tsx
```diff
@@ -33,6 +33,7 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
   const [audioSupported, setAudioSupported] = React.useState<boolean>(browserSupportsAudio);
   const [recordingStarted, setRecordingStarted] = React.useState<boolean>(false);
   const [recordingDisabled, setRecordingDisabled] = React.useState<boolean>(false);
+  const [recordingFailed, setRecordingFailed] = React.useState<boolean>(false);
   const [playDisabled, setPlayDisabled] = React.useState<boolean>(false);
   const [stopDisabled, setStopDisabled] = React.useState<boolean>(true);
```
kswenson, line 38: FYI, the `<boolean>` is unnecessary as TypeScript will infer the type correctly without it.

emcelroy: @kswenson Good to know. Thanks 👍

scytacki (changes requested):
Can student can stop the recording early? If so, it would be useful to record the length of the recording if it isn't too hard.

emcelroy:
> Can student can stop the recording early? If so, it would be useful to record the length of the recording if it isn't too hard.

The latest changes record the length of the recording in seconds (rounding up to the nearest second).

scytacki (approved):
Looks good to me.

PR #178, opened 2022-04-08
Fix snapshot content is blank

emcelroy:

[#181808855]

These changes fix a problem in runtime where the ID of a linked interactive might be set to a value that does not exist in the DOM. They add a hook that finds a linked interactive that has a specified label, and provides the correct ID value.


scytacki (approved):
This looks great to me. Nice job on the tests.

I'd be OK with merging this as is, and then we can do the graph fix and maybe remove the use of useLinkedInteractives as a separate PR.

PR #180, opened 2022-04-11
Fix snapshot content is blank

emcelroy:

[#181808855]

Adds `InitMessageContext` context and `useContextInitMessage` hook. `useContextInitMessage` should be used instead of `useInitMessage` in components that use `BaseApp` or `BaseQuestionApp`. This will limit calls to `useInitMessage` to just one which is desired since multiple calls can lead to unnecessary additional renders.

### DIFF
--- a/src/drawing-tool/components/take-snapshot.tsx
+++ b/src/drawing-tool/components/take-snapshot.tsx
@@ -45,7 +45,7 @@ export const TakeSnapshot: React.FC<IProps> = ({ authoredState, interactiveState
     <>
       {
         snapshotTarget &&
-        <button className={cssHelpers.apButton} onClick={handleSnapshot} disabled={snapshotInProgress} data-test="snapshot-btn">
+        <button className={cssHelpers.apButton} onClick={handleSnapshot} disabled={snapshotInProgress} data-testid="snapshot-btn">
           <CameraIcon className={cssHelpers.smallIcon} /> { interactiveState?.userBackgroundImageUrl ? "Replace Snapshot" : "Take a Snapshot" }
         </button>
       }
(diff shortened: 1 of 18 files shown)


src/shared/hooks/use-context-init-message.ts
```diff
@@ -0,0 +1,9 @@
+import { createContext, useContext } from "react";
+import { IInitInteractive } from "@concord-consortium/lara-interactive-api";
+
+export const InitMessage = createContext<IInitInteractive | null>(null);
```
kswenson, line 4: I would name this `InitMessageContext` to make the distinction between the message and the context that wraps it clear, and then client code would look like:
```
      <InitMessageContext.Provider value={initMessage}>
        { renderMode() }
      </InitMessageContext.Provider>
```

emcelroy: @kswenson Thanks. Makes sense to me.

src/drawing-tool/components/take-snapshot.tsx
```diff
@@ -45,7 +45,7 @@ export const TakeSnapshot: React.FC<IProps> = ({ authoredState, interactiveState
     <>
       {
         snapshotTarget &&
-        <button className={cssHelpers.apButton} onClick={handleSnapshot} disabled={snapshotInProgress} data-test="snapshot-btn">
+        <button className={cssHelpers.apButton} onClick={handleSnapshot} disabled={snapshotInProgress} data-testid="snapshot-btn">
```
scytacki, line 48: @tejal-shah is this going to break any integration tests? 
This is the snapshot button used in the image question  interactive, drawing tool interactive, and labbook interactive.

scytacki (changes requested):
This empty file seems to have been commit:
src/shared/utilities/linked-interactive-id.ts

It would be nice to change `useLinkedInteractives` so it only runs when the `initMessage.mode` is `authoring`.  This would make sure that none of the runtime code is depending on the authoring state being modified.  Based on the author changes you've made I would expect this to work.

PR #182, opened 2022-04-13
Fix snapshot content is blank -- adjust use of useContextInitMessage

emcelroy:

[#181808855]

`useLinkedIInteractives` and `useRequiredQuestion` cannot use `useContextInitMessage` since they are called by the components rendering the context (instead of by child components inside the context). So instead, we pass `initMessage` as a parameter to them.

These changes rely on changes in [this LARA pull request](https://github.com/concord-consortium/lara/pull/935) that will update the LARA Interactive API.

### DIFF
--- a/src/shared/hooks/use-context-init-message.ts
+++ b/src/shared/hooks/use-context-init-message.ts
@@ -5,6 +5,9 @@ export const InitMessageContext = createContext<IInitInteractive | null | undefi
 
 export const useContextInitMessage = () => {
   const initMessage = useContext(InitMessageContext);
+  // initMessage will be undefined unless a value is supplied by a context provider.
+  // useInitMessage in the interactive client returns null until the init message is ready,
+  // after which it returns an object of type IInitInteractive
   if (initMessage === undefined) {
     throw new Error("useContextInitMessage must be used within a <InitMessageContext.Provider>");
   }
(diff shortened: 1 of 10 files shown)


scytacki (approved):
Looks good to me.

src/shared/hooks/use-context-init-message.ts
```diff
@@ -1,9 +1,12 @@
 import { createContext, useContext } from "react";
 import { IInitInteractive } from "@concord-consortium/lara-interactive-api";
 
-export const InitMessageContext = createContext<IInitInteractive | null>(null);
+export const InitMessageContext = createContext<IInitInteractive | null | undefined>(undefined);
 
 export const useContextInitMessage = () => {
   const initMessage = useContext(InitMessageContext);
+  if (initMessage === undefined) {
```
scytacki, line 8: Does this works because useInitMessage from the interactive client returns `null` when the init message isn't ready yet?  If that is true, then a comment to that effect would be useful here.

scytacki (changes requested):

src/shared/hooks/use-context-init-message.ts
emcelroy: Yeah, as I understand it, `useInitMessage` will return `null` when the init message isn't ready. See  https://github.com/concord-consortium/lara/blob/e05bc94cbe077833384463bc9a9b3a4b54e77deb/lara-typescript/src/interactive-api-client/hooks.ts#L105

I was going to ask what you thought of this change once the checks were done. Looks like they passed and the change seems to work as expected in my testing. Does it seem reasonable to you? I'll add a comment like you suggested if so.

scytacki: @emcelroy Yes this seems reasonable.

kswenson (approved):
👍 LGTM

PR #183, opened 2022-04-13
Downgraded video.js to 7.8.4, pinned in package.json

dougmartin:


dougmartin:
Not needed, issue was fixed by converting the file from .mov to .mp4.

PR #184, opened 2022-04-15
Fix student responses missing

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181899333

[#181899333]

### DIFF
--- a/src/shared/components/iframe-runtime.tsx
+++ b/src/shared/components/iframe-runtime.tsx
@@ -25,12 +25,13 @@ interface IProps {
   setHint?: (state: any) => void | null;
   addLocalLinkedDataListener?: (request: IAddLinkedInteractiveStateListenerRequest, phone: IframePhone) => void;
   scale?: number;
+  onUnloadCallback?: (state: any) => void;
 }
 
 export const IframeRuntime: React.FC<IProps> =
   ({ authoredState, id, iframeStyling, interactiveState, logRequestData, report,
       url, setHint, setInteractiveState, addLocalLinkedDataListener, initMessage,
-      scale }) => {
+      scale, onUnloadCallback }) => {
     const [ iframeHeight, setIframeHeight ] = useState(300);
     const [ internalHint, setInternalHint ] = useState("");
     const iframeRef = useRef<HTMLIFrameElement>(null);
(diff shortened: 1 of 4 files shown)


src/shared/components/iframe-runtime.tsx
```diff
@@ -63,8 +64,12 @@ export const IframeRuntime: React.FC<IProps> =
         if (options.unloading) {
           return new Promise<any>(resolve => {
             resolveOnUnload.current = resolve;
-            // send the unload request to the wrapped interactive
-            phone.post("getInteractiveState", options);
+            if (!isSubinteractive) {
```
scytacki, line 67: This approach isn't great, but something like it might be the best quick fix.

It isn't great because it means there might be lost work when the student changes the page immediately after making a change in the sub interactive. 

A better, but harder fix, would be for the hosting interactive to provide the "setOnUnload" function as a prop to the iframe-runtime. Then the hosting interactives can take the result from `resolveOnUnload.current?.(newInteractiveState);` and wrap it the same way they are wrapping the interactive state of their sub interactives normally.

scytacki: cc @dougmartin

emcelroy: @scytacki Thanks. Yeah, I kind of suspected it wouldn't be this easy or it would have been done already. I put this up so I could show Doug when I asked him (on Slack) if it'd be a bad idea.

I'll look into the approach you suggested.

src/shared/components/iframe-runtime.tsx
```diff
@@ -69,10 +67,17 @@ export const IframeRuntime: React.FC<IProps> =
...
-        resolveOnUnload.current?.(newInteractiveState);
+        if (onUnloadCallback && resolveOnUnload.current) {
+          // send the interactive state to any parent interactive that has provided an onUnload 
+          // callback, and then resolve the promise that was saved in the setOnUnload function 
+          // with undefined so the parent interactive doesn't have its state overwritten.
+          onUnloadCallback(newInteractiveState);
+          resolveOnUnload.current(undefined);
+        } else {
+          resolveOnUnload.current?.(newInteractiveState);
+        }
```
dougmartin, line 80: This is not really about this change but I'm wondering though if we should set `resolveOnUnload.current` to `null` here so that we are guaranteed to only call it once.

dougmartin (approved):
Looks good :+1:   My mental runtime says it should work :)

src/shared/components/iframe-runtime.tsx
emcelroy: @dougmartin Thanks. I've made that change and it seems to work fine. Look good to you?

dougmartin (changes requested):
Switch to setting the current to null and not returning null (and return undefined instead).

PR #185, opened 2022-04-15
Upgrade to lara-interactive-api 1.6.0

dougmartin:
This is required to enable the onUnload interactive state hook


knowuh (approved):
Huge review :D

PR #189, opened 2022-04-20
Fix snapshot content is blank

emcelroy:

[#181808855]

Remove the Side-by-Side Interactive's setting of `onUnloadCallback` because it is not needed and it is causing infinite re-rendering of child components.

Note that the re-rendering issue was identified in snapshot bug QA, but it was caused by changes made to fix the lost student response bug (PT Story: https://www.pivotaltracker.com/story/show/181899333).


dougmartin (approved):
Looks good :+1:

PR #190, opened 2022-05-02
Open Response Record/Play/Stop/Delete/Time Controls

emcelroy:
PT Stories: 

https://www.pivotaltracker.com/story/show/181573717
https://www.pivotaltracker.com/story/show/181888817
https://www.pivotaltracker.com/story/show/181866053

[#181573717]
[#181888817]
[#181866053]

These changes improve the runtime UI for recording open response audio by increasing the button size for better accessibility and adding a time readout for recording and playback of audio. Also included are fixes for bugs related to the placement of the audio controls. The controls will now automatically move to the appropriate location in the textarea if the latter is resized, and text within the textarea will never overlap with the controls because the field will automatically increase in height to leave space for the controls no matter how much text is entered.


scytacki (approved):
It looks good to me. I didn't review as carefully as I do sometimes, so you might want to get a second opinion. :)

dougmartin (approved):
Looks good :+1:

PR #191, opened 2022-05-13
Bug Fixes for Record/Play/Stop/Delete/Time Controls

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181573717

[#181573717]

Fixes some bugs reported by QA.

1) Clicking on the record button while recording was in progress was re-triggering `handleAudioRecord` even though the button was disabled. So now we set the button's `onClick` property to `undefined` while recording is in progress.

2) Component wasn't fully wrapping up the audio recording process when user navigated to another page before manually stopping recording. So now we use `handleAudioRecordStop` instead of just `mediaRecorderRef.current.stop()` to ensure the process is fully completed. This makes sure that the recording timer doesn't keep running when recording is automatically stopped.

3) The first time a user leaves an audio response but no text response to a question and then navigates to another page, the error "Some items on this page did not save" would appear. So now in addition to checking if there's a media recorder on unload, we check if recording is in progress. If recording is not in progress, we resolve with the interactive state.


scytacki (approved):
Seems fine to me.

PR #192, opened 2022-05-17
Bug Fix for Record/Play/Stop/Delete/Time Controls

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/181573717

[#181573717]

These changes ensure that user cannot navigate away from page until an active audio response recording has stopped AND successfully saved.


scytacki (approved):
looks good

PR #193, opened 2022-05-23
Adjusting layout so labbook does not get cut off in 40-60 layout.

lublagg:


src/labbook/components/thumbnail-chooser/thumbnail-chooser.tsx
```diff
@@ -14,7 +14,7 @@ interface basicButtonProps {
 
 const PrevButton:React.FC<basicButtonProps> = (props: basicButtonProps) => {
   const {onClick, enabled} = props;
-  const backClasses = classNames(css["button-back"], {[css.disabled]: !enabled});
+  const backClasses = classNames(css["arrow-button"], css["back"], {[css.disabled]: !enabled});
```
knowuh, line 17: I didn't see "back" or "forward" classes in `./thumbnail-chooser.scss` -- Not sure if the class name is used in tests or somewhere else, but maybe we can remove this.

I think originally the name `button-back` came from the Zeplin specs. This might be a naming convention Michael uses when talking about the background div in which an SVG button sits.  I see here it's use was confusing because of the forward and backward navigation.

knowuh (approved):
Thank you Lucy,

The code looks good to me.

I am not 100% sure it is solving the problem, but its probably an improvement. 

 I have attached screenshots of what it looks like when I run this in the activity player, launched from authoring.staging.

Screen Shot 2022-05-25 at 6 16 28 PM
Screen Shot 2022-05-25 at 6 13 39 PM
Screen Shot 2022-05-25 at 6 16 59 PM

src/labbook/components/thumbnail-chooser/thumbnail-chooser.tsx
lublagg: You're right, I was messing around with adding those classnames then forgot to delete them. I just removed them & committed.

In terms of the screenshots above, I made the area with the snapshots horizontally scrollable, so even though it looks like you can't see all of them, you can actually. Maybe that's not clear enough, though.

src/labbook/components/runtime.tsx
```diff
@@ -61,6 +63,18 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
...
   const {maxItems, showItems} = authoredState;
   const {entries, selectedId} = ensureSelected(interactiveState as IInteractiveState) as IInteractiveState;
+  const containerRef = useRef<HTMLDivElement>(null);
+
+  // checking if app container is smaller than full-width; if so, we need to scale down app
+  useEffect(
+    () => {
+      if (containerRef.current && containerRef.current.clientWidth < 514) {
+        setContainerWidth(containerRef.current.clientWidth);
+        setCanvasWidth(containerRef.current.clientWidth - 42);
```
kswenson, line 73: More examples of magic numbers that could benefit from named constants.

knowuh (approved):
I still approve of this PR.

@kswenson had some good comments, but if you just want to merge as is, I can support that.

I think it would be quick to address his concerns though.

PR #194, opened 2022-06-02
Fixed carousel image jitter [#182311084]

dougmartin:
This adds an optional scrolling prop to the shared iframe runtime component.  The carousel runtime sets this scrolling prop to "no" to fix the image jitter.

### DIFF
--- a/src/shared/components/iframe-runtime.tsx
+++ b/src/shared/components/iframe-runtime.tsx
@@ -26,7 +26,7 @@ interface IProps {
   addLocalLinkedDataListener?: (request: IAddLinkedInteractiveStateListenerRequest, phone: IframePhone) => void;
   scale?: number;
   onUnloadCallback?: (state: any) => void;
-  scrolling?: string;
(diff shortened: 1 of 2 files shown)


src/shared/components/iframe-runtime.tsx
```diff
@@ -26,12 +26,13 @@ interface IProps {
   addLocalLinkedDataListener?: (request: IAddLinkedInteractiveStateListenerRequest, phone: IframePhone) => void;
   scale?: number;
   onUnloadCallback?: (state: any) => void;
+  scrolling?: string;
```
kswenson, line 29: Perhaps:
```suggestion
  scrolling?: "auto" | "yes" | "no";
```

dougmartin: @kswenson Good point.  Fixed in https://github.com/concord-consortium/question-interactives/pull/194/commits/a07903f6b5b8f8346659fc71ec167ec10043789a.

kswenson (approved):
👍 Looks good

PR #196, opened 2022-06-03
Align Lato font link with AP [#180588632]

dougmartin:
AP uses 400, 700, and 900 font-weight for Lato.  This change makes question-interactives align with that plus adds 100 as locked-info.scss uses lighter for a font-weight which is 100 when at normal 400 weight.

This also sets the $default-font-weight SCSS variable to 400.  Currently this variable is not used but may be used in the future.


dougmartin:
@knowuh / @scytacki - this fixes the bold font rendering on Macs (it looked fine on Linux).  Previously only the 400 weight font was being downloaded - now all the font weights used are being downloaded.  First one to review wins a lollipop.

knowuh (approved):
Looks good to me

PR #197, opened 2022-06-15
Added optional flushOnSave to iframe-runtime [#182039050]

dougmartin:
This is used in the full-screen interactive to flush the saves immediately so that linked interactive previews can access the latest save data.


scytacki (approved):
Looks fine to me.

Tests of this would be worthwhile, but not required.

PR #199, opened 2022-06-17
Fix Drag & Drop items not registering if placed on top of another item.

emcelroy:
PT Story: https://www.pivotaltracker.com/story/show/182015941

[#182015941]

When an item is being dragged, we should disable all other draggable items until the item being dragged is dropped. This ensures that items dropped into a drop zone will register as dropped even if they are dropped directly on top of another item that's already in the drop zone.

### DIFF
--- a/src/drag-and-drop/components/container.tsx
+++ b/src/drag-and-drop/components/container.tsx
@@ -192,7 +192,7 @@ export const Container: React.FC<IProps> = ({ authoredState, interactiveState, s
     }
   }, [setInteractiveState]);
 
-  const [, drop] = useDrop({
+  const [{ itemDragging, isOver }, drop] = useDrop({
     accept: [DraggableItemWrapperType, DropZoneWrapperType],
     drop(wrapper: IDraggableItemWrapper | IDropZoneWrapper, monitor) {
       const didDrop = monitor.didDrop();
@@ -225,6 +225,7 @@ export const Container: React.FC<IProps> = ({ authoredState, interactiveState, s
       }
     },
     collect: monitor => ({
+      itemDragging: monitor.getItem(),
       isOver: !!monitor.isOver(),
     }),
   });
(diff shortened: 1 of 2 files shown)


src/drag-and-drop/components/container.tsx
```diff
@@ -275,7 +276,11 @@ export const Container: React.FC<IProps> = ({ authoredState, interactiveState, s
               top: Math.min(canvasHeight - margin, top)
             };
           }
-          return <DraggableItemWrapper key={item.id} item={item} position={position} draggable={!readOnly} />;
+          // When an item is being dragged, temporarily disable all other draggables until it is dropped. This 
+          // ensures the item registers as dropped in a drop zone even if it is dropped on top of another 
+          // draggable that is already in the drop zone.
+          const tempDisable = isOver && itemDragging.item.id !== item.id;
```
kswenson, line 282: Out of curiosity and to make sure I'm understanding things, wouldn't it also work to eliminate the `isOver` test, i.e. to always disable items other than the one being dragged, whether or not we're currently over them?

kswenson (approved):
👍 LGTM -- just one curiosity question

src/drag-and-drop/components/container.tsx
emcelroy: @kswenson `isOver` is coming from [`useDrop`](https://react-dnd.github.io/react-dnd/docs/api/use-drop) which means it's describing whether a draggable is over the drop zone, not whether it's over another draggable. So the draggables not being dragged are always disabled when an item is being dragged over the drop zone. 

I probably should've named the variable something like `isOverDropZone` to make that clear.

I had tried not checking `isOver` at one point, but the results were strange. Trying it out again, though, it does seem to work without it. I must've been doing something else wrong when I tried that before.

I've dropped the use of `isOver` in the latest commit. Does that look good?

kswenson (approved):
👍 Looks good -- simple and to the point!

PR #203, opened 2022-07-01
ScoreBOT: user report item API to generate feedback HTML

pjanik:
Two main changes here:
1. `renderStyledComponentToString` helper has been added. It's based on @knowuh's implementation used in Geocode. That obviously required installing `styled-components`.
2. ScoreBOT uses this helper to render its report feedback component to string. Some components had to be refactored to use styled-components.
3. Random cleanup and refactoring in ScoreBOT. I've extracted helpers to a separate file and removed `score` from the interactive state top-level (it was redundant, I can use the last attempt score).
4. TODO: unit tests that cover better ScoreBOT-specific features.

Example 1:
Screenshot 2022-07-01 at 21 21 10
Example 2:
Screenshot 2022-07-01 at 21 19 53


knowuh (approved):
Looks good,

Thank you!

PR #204, opened 2022-07-05
182585954 fix rte dialog width

eireland:
Updates slate-editor library to version 0.8.0
Adds LARA specific styling to slate editor modals into LARA so it fits in LARA modals.
Removes `order` from SlateToolbar props as it no longer exists in 0.8.0


emcelroy (changes requested):
This looks good except there are elements within the dialog that need to have their widths adjusted, too. For example:

image

It still seems weird to me that we're hard coding specific widths into the rich text editor instead of setting elements' width to `100%` or `auto` and then setting a `max-width` of whatever width is considered optimal. I think that may be a simpler way to get the same result, but maybe there's some reason that won't work that I'm missing.

eireland:
I tried that, but it caused positioning to be wrong. The simplest was to specify it in question interactives.

emcelroy (changes requested):
The CSS change looks good. The automated checks are failing, though. I think you have to commit the updated package-lock.json file.

`npm ERR! cipm can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.`

emcelroy (approved):
Looks good -- thanks 👍

PR #207, opened 2022-07-11
Return "score" report item from ScoreBOT interactive, refactor report-items to use `useReportItem` hook

pjanik:
This PR is related to the new Interactive API: https://github.com/concord-consortium/lara/pull/982

Also, I've refactored report items to use a new hook. I hope it makes it a bit simpler, we don't need to manually notify the host that report item is ready.

Once this approach is approved, I'll add more tests to increase the coverage.


scytacki (approved):
Looks good to me. (with the exception of the test coverage)

PR #208, opened 2022-07-13
Add report items tests

pjanik:
This PR adds missing report item tests, both for Open Response and ScoreBOT.
I've slightly reorganized code between app.tsx and report-item.tsx files. report-item now only includes the handler function, so it's easier to write unit tests. I've realized that `useReportHandler` hook could make that even simpler by calling `sendReportItemAnswer` itself and expecting the provided handler to return `items` only. On the other hand, it'd make the the logic more hidden and it might be not so easy to understand what's going on in the report item. So, I've left that for now, as it's a minor change anyway.

### DIFF
--- a/src/score-bot/components/report-item/report-item.tsx
+++ b/src/score-bot/components/report-item/report-item.tsx
@@ -18,8 +18,6 @@ export const reportItemHandler: IGetReportItemAnswerHandler<IInteractiveState, I
     const items: IReportItemAnswerItem[] = [];
 
     if (itemsType === "fullAnswer") {
-      // This item intentionally doesn't include answer text. Answer text is obtained by portal report itself
-      // and preloaded while it's waiting for report items response.
       items.push({ type: "answerText" });
 
       const score = getLastScore(interactiveState);
--- a/src/score-bot/components/report-item/report-item.tsx
+++ b/src/score-bot/components/report-item/report-item.tsx
@@ -1,67 +1,52 @@
 import * as React from "react";
 import * as semver from "semver";
-import {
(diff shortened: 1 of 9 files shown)


src/score-bot/components/report-item/report-item.tsx
```diff
@@ -1,67 +1,52 @@
...
+    // tslint:disable-next-line:no-console
+    console.error("ScoreBOT Report Item Interactive: Missing version in getReportItemAnswer request.");
+  }
+  else if (semver.satisfies(version, "2.x")) {
+    const items: IReportItemAnswerItem[] = [];
+
+    if (itemsType === "fullAnswer") {
+      // This item intentionally doesn't include answer text. Answer text is obtained by portal report itself
+      // and preloaded while it's waiting for report items response.
+      items.push({ type: "answerText" });
```
scytacki, line 23: this comment is confusing. It says the item doesn't include answer text and then adds the answerText item.  Perhaps the intention is that `html` item that is added doesn't include the answer text since it will be added based on the `answerText` item.

scytacki (approved):
This looks good to me.

I just found one comment from earlier code that was confusing .

src/score-bot/components/report-item/report-item.tsx
pjanik: So, what I meant in this comment is that I'd typically expect this line to look like this:
```
items.push({ type: "answerText", answerText: interactiveState.answerText });
```
but instead, we push an "empty" item (just type, no content):
```
items.push({ type: "answerText" });
```
I wanted to clarify that the content of the answer text is calculated by the portal-report itself, as it was confusing for me at the beginning (I think we discussed that on slack one day).

So, any suggestion for better wording here so the comment doesn't make it even worse? :)

PR #211, opened 2022-07-18
Changes width of iframe authoring subsequent questions

eireland:
Changes width of iframe authoring in carousel so image editing dialog and rte toolbar are visible.
Changes here need to be viewed withcorrespoonding chnages in LARA PR in https://github.com/concord-consortium/lara/pull/990


emcelroy (approved):
Looks good 👍

PR #212, opened 2022-07-20
182585954 rte dialog height fix

eireland:
removes top and bottom margins to labels so that dropdown in image edit dialog in fill in the blank authoring in carousel question interactive is fully visible.
Note that header is cutoff and should be addressed in the future.


emcelroy (approved):
This helps with accessing the drop-down and it seems reasonable to minimize those margins. There seems to be a larger problem with the dialog box being cropped and partially inaccessible, though, that seems like it requires a more comprehensive fix. I guess that'll be part of figuring out the problem with the header being cut off?

PR #214, opened 2022-07-26
Add a script to convert LARA answers to AP Managed Interactive answers

pjanik:
[#182525578]

This early version of the script has some unnecessary logic / items marked with TODO comments. It lets me run it on the production data before the migration is done and measure performance.

I've added unit tests to double-check the results and clearly see what we're copying around. Typing is also probably overspoken, but it was done so it's easier to know what data I was working with.

Performance varies a lot, depending on activity structure. It obviously becomes the slowest when there's a sequence with plenty of answers (we have some that have 3mln and more). For example:

```
Elapsed time: 590.42 seconds
Processed 1000 activities, 1.69 activities per second
Processed 11105 questions, 18.81 questions per second
Copied 1928873 answers, 3266.96 answers per second
Max number of answers per question: 9543
```
vs
```
Elapsed time: 996.12 seconds
Processed 10 activities, 0.01 activities per second
Processed 275 questions, 0.28 questions per second
Copied 4342234 answers, 4359.15 answers per second
Max number of answers per question: 28898
```

But the answer conversion rate is consistent. Still, the big unknown is how many answers we have to deal with.

Next, I'll add more configuration, so I can run the script for logged-in or anonymous users, process activities only within some dates, and maybe even process sequences separately. As they'll have significantly more answers per sequence, so I could probably configure batching in a different way. Right now it's very conservative to avoid issues with Firestore and memory.


src/convert-lara-answers/convert-interactive-state.test.ts
```diff
@@ -0,0 +1,147 @@
...
+      tool_user_id: '1406',
+      id: 'converted-authoring.concord.org-answers-interactive_run_state_78721',
+      answer: '{"version":1,"mode":"report","authoredState":null,"interactiveState":"{\\"data\\":[[\\"a\\",\\"Table\\",\\"that\\",\\"saves\\"],[\\"b\\",null,\\"state\\",null],[\\"c\\",null,null,null]]}"}',
+      type: 'interactive_state',
+      resource_link_id: '2462',
+      remote_endpoint: 'https://learn.staging.concord.org/dataservice/external_activity_data/933e2f3f-13b2-4119-b3b9-97485aaa8c58',
+      tool_id: 'activity-player.concord.org',
+      class_info_url: 'https://learn.staging.concord.org/api/v1/classes/32',
+      run_key: '19bfcb93-ccbf-4f53-84a7-296294821954',
+      question_id: 'managed_interactive_448816',
```
scytacki, line 88: The question_id is changed here from 'mw_interactive_212649'.
I would have expected it to stay the same since we are not converting existing interactives. 
But I guess it is just because of the mocking of the `new_question`.

scytacki (approved):
This looks good to me.

If I understand it right, we can run it over every logged in answer first and then run it over every anonymous answer. I think that will be a good enough way to segment it. I don't think we need the date filtering, but we can discuss more.

If you make significant changes, you can re-request my review so I know to look at the changes.

src/convert-lara-answers/convert-interactive-state.test.ts
pjanik: Yes, in the real-world scenario, the passed question would be an old question in fact. So, the ID will be updated to the same value.

pjanik:
> If I understand it right, we can run it over every logged in answer first and then run it over every anonymous answer.

Right.

> I think that will be a good enough way to segment it. I don't think we need the date filtering, but we can discuss more.

I made that optional. I thought about running 2022 first.

Also, I've just requested another review, as there are plenty of changes. The script:
- will ignore failed answer conversion (but log the error in error log file, and answer id in a separate json, so we can check what happened and even inspect answers manually in Firestore)
- try to save batch of converted answer up to 5 times, waiting between attempts (fighting with Firestore write limit)
- try to process each resource up to 5 times, waiting between attempts (fighting with Firestore write limit)
- all the errors should be handled at some level (answer conversion, answer batch saving, resource processing) and the script should continue to run no matter what happens - I've tried to crash it in multiple ways
- if a resource still cannot be successfully processed after 5 attempts, it'll be logged in a separate JSON file, so we can later process them again
- logging is improved (separate log files for progress, errors, info, and failed resources and answers. Some of them are valid JSONs so it's easier to process them by another script if necessary)
- + other misc changes

scytacki (approved):
Looks good to me.

PR #215, opened 2022-08-01
Add question feedback conversion script

pjanik:
This PR adds a feedback conversion script. It also includes some changes to the main answers conversion script, but these have been already used during the main production data copying (I've added a few more fields and tweaked some params).

I've tested feedback locally on my activities and it seems to work fine. There's one issue - teachers will see that their feedback for MC question is outdated. This happens, as the feedback code compares the answer hash with the answer hash that it's saved in the feedback doc. It also means that students won't see the feedback in their reports (as the report doesn't show outdated feedback to students).

Screenshot 2022-08-01 at 22 34 35
Screenshot 2022-08-01 at 22 34 27
(these MC questions should show a regular green rectangle)

It happens as we changed MC answer format (string IDs instead of numbers). I can try to fix that tomorrow.


scytacki (approved):
Looks good to me

pjanik:
It seems that feedback is a very low priority, so I'm not gonna play with recalculating the MD5 hash and leave this issue as it is. If we ever need to fix it, we should still be able to do it.

PR #216, opened 2022-08-04
only show the text area when answer prompt is set

scytacki:
For consistancy this also hides any student written answer if
the answer prompt is not set.

Additionally this PR:
- switches to webpack serve, this supports https development.
- updates lint rules so webpack and eslint pass
- update webpack.config.js to use single quotes


pjanik (approved):
Looks good to me. 👍

PR #217, opened 2022-08-05
ScoreBOT: truncate score if necessary

pjanik:
This PR provides an ugly workaround for the issue described here:
https://www.pivotaltracker.com/n/projects/2441249/stories/182902508
It's possible that LARA implementation was doing similar (explicitly or somewhat hidden).
I think it's better to do it here rather than in the API and add more confusion there. Also, this workaround won't break anything if we fix the API, so we don't need to remove it. 

@scytacki, I'm providing this "fix", as Leslie said she talked with you about it and it seemed like a reasonable solution for now.

### DIFF
--- a/src/score-bot/components/runtime.tsx
+++ b/src/score-bot/components/runtime.tsx
@@ -25,6 +25,12 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
 
   const decorateOptions = useGlossaryDecoration();
 
+  const maxScore = getMaxScore(authoredState);
+  const lastScore = getLastScore(interactiveState);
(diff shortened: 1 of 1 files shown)


scytacki (approved):
Looks good to me.

I'd guess that the use of appending the itemID into the answer and using one model never really worked right.  It is only giving the model one number to differentiate the answers by. 

If we come back to this, a better approach would be to include the full prompt of the question in the answer text used by the model. That would hopefully be enough context for the model to differentiate between answers to different questions. And additionally if the target of the model could be two outputs: a score and question id, instead of just the score. That ought to help the model learn which question the answer is for as well as help us evaluate how the model is doing with differentiating answers.

PR #218, opened 2022-08-15
Fixed submit button disabled when there is no answer prompt [#182914990]

dougmartin:


src/image-question/components/app.tsx
```diff
@@ -49,7 +49,12 @@ const baseAuthoringProps = deepmerge(drawingToolBaseAuthoringProps, {
   arrayMerge: (destinationArray, sourceArray) => sourceArray
 });
 
-const isAnswered = (interactiveState: IInteractiveState | null) => !!interactiveState?.answerText;
+export const isAnswered = (interactiveState: IInteractiveState | null, authoredState?: IAuthoredState) => {
+  const hasAnswerPrompt = !!authoredState?.answerPrompt;
+  const hasAnswerText = !!interactiveState?.answerText;
+  // image questions without answer prompts are always answered otherwise they are answered when there is answer text present
+  return !hasAnswerPrompt || hasAnswerText;
```
scytacki, line 56: They should require some kind of work in the drawing area. This could be actual drawing, or it could just be an snapshot or uploaded image.  Perhaps just having `interactiveState` is enough to determine this.

scytacki (changes requested):
Previously this code would also say the interactive was unanswered if the interactive state was null.  I think it will now say it is answered in this case.

dougmartin:
> Previously this code would also say the interactive was unanswered if the interactive state was null. I think it will now say it is answered in this case.

It is using the old code plus a check for the answer prompt now.

The old code code was:  `!!interactiveState?.answerText;` and the new code (reduced to remove the variables) is `!authoredState?.answerPrompt || !!interactiveState?.answerText`

src/image-question/components/app.tsx
dougmartin: So you want me to change the answer check altogether? Previously it was just checking the answerText.

scytacki:
Right. I'm pretty sure with that logic if interactiveState is null in the old code it would return false. 
In the new code if answerPrompt is not present it will always return true.
So in the new code if an author doesn't add an answerPrompt the submit button will be enabled even before the student does anything with the question.

dougmartin:
> Right. I'm pretty sure with that logic if interactiveState is null in the old code it would return false.

Got it. Fixed in https://github.com/concord-consortium/question-interactives/pull/218/commits/aca8ec60f22a2c541621a45d401ff71366e724b4.

scytacki (approved):
Looks good!

PR #219, opened 2022-08-16
Updated drawing-tool to remove "ngsa" from "ngsaObject" in the stamp collection name [#182908556]

dougmartin:
This removes "ngsa" from the authoring stamp collection option and removes it from the default name at runtime when the stamp is loaded with no name override.


scytacki (approved):
seems fine to me

PR #228, opened 2022-10-12
Upgraded and integrated ChartJS into bar chart [#183479133]

dougmartin:
This upgrades chart.js, chartjs-plugin-datalabels and react-chartjs-2 to the latest version and updates the existing graph interactive to use it along with integrating it with the bar-chart interactive.


lublagg (approved):
Looks good to me.

PR #230, opened 2022-10-13
Update Labbook not to use Shutterbug

pjanik:
[#183480687]

This PR makes a few changes, but the goal is not to take any snapshots inside the Labbook interactive. This wasn't really necessary. Labbook can use a small Drawing Tool instance in its thumbnail to display the same state that it already stores.

To make that possible:
- DrawingTool library had to be updated to version 2.2.0, so it supports canvas-only mode (no buttons panel, no canvas styling) and canvasScale (see: https://github.com/concord-consortium/drawing-tool/pull/63)
- DrawingTool interactive had to be updated, so multiple Drawing Tools can be instantiated on the same page (before it was using hardcoded DOM ID)
- Drawing Tool read-only overlay has been re-implemented, as it had been capturing thumbnail clicks
- Drawing Tool read-only overlay is now optional (its visual aspect), as we don't want that in the thumbnail
- Drawing Tool accepts a new `canvasScale` param passed to a Drawing Tool library
- Drawing Tool now notifies the parent about state update once the background is set (so user can immediately see snapshot in the thumbnail)
- Labbook interactive has been updated to use the new Drawing Tool in its thumbnail (+ some styling changes so thumbnail content fits the available area better)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "deep-equal": "^2.0.5",
         "deepmerge": "^4.2.2",
         "dompurify": "^2.3.6",
-        "drawing-tool": "^2.1.2",
+        "drawing-tool": "^2.2.0-pre.1",
         "eslint-webpack-plugin": "^3.1.1",
         "firebase": "^9.6.8",
         "html-react-parser": "^0.13.0",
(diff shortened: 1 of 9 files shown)


dougmartin (approved):
Looks good :+1:

PR #232, opened 2022-10-14
Labbook: add "Show Upload Image button" authoring option and don't show "Take Snapshot" button when not necessary

pjanik:
[#182021867]

This PR adds "Show Upload Image button" authoring option to Labbook. It also fixes "Take Snapshot" button that was always displayed, even if it wasn't functional at all (when the background source was URL or Upload).


dougmartin (approved):
Looks good :+1:

PR #234, opened 2022-10-17
Improved bar-chart accessibility [#183562465]

dougmartin:
Moved title, subtitle and x-axis out of chartjs rendered canvas and into DOM.  Added title attribute to bar value that renders "<bar name>: <bar value> <x axis?".


lublagg (approved):
Looks good!

PR #242, opened 2022-10-19
Image Question: fix URL parsing

pjanik:
[[#183580877]](https://www.pivotaltracker.com/n/projects/2441249/stories/183580877)

The previous approach wasn't working when the URL didn't have any search params.


dougmartin (approved):
Looks good :+1:

PR #245, opened 2022-10-20
Added "thumbnail" report-item view of the graph [#183479609]

dougmartin:
Also updated the demo component to optionally display the report item html if configured to do so.


lublagg (approved):
Looks good!

PR #255, opened 2022-10-27
Labbook: fix Upload button when background source is set to URL

pjanik:
I think this story and comments provide the best context: https://www.pivotaltracker.com/story/show/183592738

Generally, Labbook always shows the "Upload" button, even if the background source is set to URL. But the user-uploaded background was later ignored by the DrawingTool component. This PR fixes that.


dougmartin (approved):
Looks good :+1:

PR #256, opened 2022-10-27
Workaround LARA linked interactive issue

pjanik:
[[#182710781]](https://www.pivotaltracker.com/story/show/182710781)

I'm opening this PR only in case the bug is important to fix ASAP. I doubt so, as it's been around for a while (it's not a regression, but most likely LARA2 problem). It's also a small note to myself about the problem. Otherwise, I wouldn't merge it.

Generally, ManagedInteractive.to_hash method returns a different form of linked_interactives than what the update request is expecting. And there seems to be an assumption that they're the same. Unfortunately, I don't know how to fix that properly in LARA yet. I'll have to discuss it with Ethan or Scott when I'm back next week.


pjanik:
This has been fixed properly by https://github.com/concord-consortium/lara/pull/1036, so I'm closing this PR.

PR #259, opened 2022-11-04
Fix image upload

pjanik:
[[#183721375]](https://www.pivotaltracker.com/story/show/183721375)


dougmartin (approved):
Looks good :+1: 

Curious as to why you just didn't add this at the end of the `onChange` handler though.

pjanik:
> Curious as to why you just didn't add this at the end of the onChange handler though.

I guess it'd work just fine too. I was kinda afraid this would interfere with the file reference that we're dealing with, especially since the upload action is async, and we'd set the value to null almost immediately. But most likely this is just my imagination. :wink:

PR #260, opened 2022-11-04
Fix Drawing Tool state saving on background update

pjanik:
Fixes:
- [**[Lab book]** Lab book is considered to be an answered question in the reports even if the user doesn't use it (opening AP page is enough)](https://www.pivotaltracker.com/story/show/183716771)
- [**[Drawing Tool]** If question is marked as required, submit button is enabled by default in version V1.10.3](https://www.pivotaltracker.com/story/show/183661940)


dougmartin (approved):
Looks good :+1:

PR #261, opened 2022-11-04
Fix Drawing Tool state observing

pjanik:
This PR fixes a bug found by Saravanen:
https://www.pivotaltracker.com/story/show/183725749
More details in another PT story:
https://www.pivotaltracker.com/story/show/183512066/comments/233988238

State update was entering an infinite loop.


dougmartin (approved):
Looks good :+1:

PR #262, opened 2022-11-04
Implement Labbook isAnswered function

pjanik:
https://www.pivotaltracker.com/story/show/183641693

Apparently, this was never implemented and it was causing Submit button to be rendered incorrectly.


dougmartin (approved):
Looks good :+1:

PR #265, opened 2022-12-02
ImageQuestion: reset history after background is set

pjanik:
This PR fixes the issue that is well described here: [[#183750971]](https://www.pivotaltracker.com/story/show/183750971)

Instead of pausing the history, it's reset right after the background image is set.

This prevents students from undoing/removing authored background. It also works better with snapshots or background uploaded by users, as they can no longer remove it using undo button - it wasn't working anyway, as image question stores this student background separately and sets it each time the interactive is initialized.


dougmartin (approved):
Looks good :+1:

PR #270, opened 2022-12-16
Update dependencies

pjanik:
[#184017470]

Dependencies update. I did that separately from monorepo conversion while we still had one package.json and extracted to a separate branch.

I've updated React to v17. It won't be easy to update to v18, as I think we might be blocked by Enzyme tests. This lib seems to be dead and doesn't (reliably?) support React v18. We'd need to convert them to react-testing-library.


scytacki (approved):
looks good to me

PR #271, opened 2022-12-16
Refactor to monorepo

pjanik:
[#184017470]

This is the current state of the conversion. The plan is to discuss it live first, but I might prepare some summary on Monday.

[EDIT]:
I've updated README and eslint-plugin-import, so I think it's ready to be merged.


pjanik:
@scytacki, in the last commits I've fixed LARA Interactive API linking scripts (I've actually had to use them and noticed this part wasn't converted well) + merged master into this branch, and resolved simple conflicts (versioning after 1.10.4 release).

scytacki (approved):
This looks good to me. I made a quick typo fix commit.

PR #274, opened 2022-12-23
TecRock Table: save interactive state to show itself in report 

pjanik:
[#184027397]

Now TecRock table will simply clone interactive state. This lets report consider this question as answered and also the table can be rendered there.


dougmartin (approved):
Looks good :+1:

PR #278, opened 2023-01-11
Fix background scroll in bar-graph [#184195561]

dougmartin:
This fixes the background scrolling on an iPad when the user drags the bar adjustment icons


lublagg (approved):
Looks good!

PR #280, opened 2023-01-12
Remove TecRock Table interactive (it was moved to tectonic-explorer monorepo)

pjanik:
[#184185704]


dougmartin (approved):
Looks good :+1: 

Too bad Noah isn't still around - his is the kind of PR he loved - all deleted files :)

PR #281, opened 2023-01-26
184171460 fix dt undo redo

pjanik:
[[#184171460]](https://www.pivotaltracker.com/story/show/184171460)
[[#184171389]](https://www.pivotaltracker.com/story/show/184171389)

This PR is related to a fix in the drawing tool library: https://github.com/concord-consortium/drawing-tool/pull/76
It was preventing correct state saving when user was using undo and redo.

I've also renamed pacakges/drawing-tool to packages/drawing-tool-interactive to avoid name conflict. I had issues to `npm link` drawing-tool library that had the same name.


dougmartin (approved):
Looks good :+1:

PR #282, opened 2023-01-26
Image: fix zoom button layout, rm unused allowLightbox

pjanik:
[[#183726314]](https://www.pivotaltracker.com/story/show/183726314)

Issues are described well in the PT story. This PR fixes zoom-in button positioning and removes unused authoring property.

### DIFF
--- a/packages/image/src/components/runtime.scss
+++ b/packages/image/src/components/runtime.scss
@@ -1,49 +1,62 @@
 @import "@concord-consortium/question-interactives-helpers/src/styles/helpers";
-.runtime{
-  height:100%;
-}
-.imageContainer{
-  width: 100%;
-  margin-bottom: 10px;
-  img{
-    max-width: 100%;
-    height: auto;
-  }
-  &.fitWidth{
-    img{
-      width: 100%;
+
+.runtime {
+  height: 100%;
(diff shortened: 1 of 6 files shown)


packages/image/src/components/runtime.scss
```diff
@@ -1,49 +1,62 @@
...
-}
 
-.credit{
-  font-size: 0.8em;
-  width: 100%;
-  line-height: 1.2em;
-}
+  .captionAndZoomIn {
+    display: flex;
+    flex-direction: row;
```
dougmartin, line 30: Do we need to add `align-items: flex-start` here so the two columns (`textContainer` and `viewHighRes`) are aligned at the top or is that already taken care of in the default flex layout?

dougmartin (approved):
Looks good :+1:   I just had a question about alignment but it isn't blocking a merge.

packages/image/src/components/runtime.scss
pjanik: It seems to work fine without it, so I guess it might be the default behavior already.
Screenshot 2023-01-26 at 19 36 23
Screenshot 2023-01-26 at 19 35 08

PR #283, opened 2023-02-01
Fix "Required" setting handling in Carousel, optimize build scripts

pjanik:
[[#184171535]](https://www.pivotaltracker.com/story/show/184171535)

This PR fixes the "Required" setting handling in Carousel.
Also, since I had to work with all the packages/interactives, I've updated build scripts to be efficient to work with (faster, no caching).


emcelroy (approved):
Looks good!

PR #285, opened 2023-02-03
Upgrade react-jsonschema-form so authoring validation works with React v17

pjanik:
[[#184374529]](https://www.pivotaltracker.com/story/show/184374529)

I've noticed that question interactives no longer show any errors in authoring. Some of them have authored state validation defined. Eg drawing tool checks if the background URL is correct and whether it uses [concord.org](http://concord.org/) domain. But this is no longer visible in the UI.

I've done some investigation using `git bisect`, and it seems this got broken when I updated React from v16 to v17 in question-interactives repo. We're using a pretty old version of react-jsonschema-form library - v1.8.1. 

So, it was a good motivation to work on this library upgrade. Fortunately, it also fixed the validation issue.


scytacki (approved):
looks good to me.

PR #286, opened 2023-02-03
feat(Carousel): add authoring form validation

pjanik:
[[#184392029]](https://www.pivotaltracker.com/story/show/184392029)

This should help authors to understand how to make the carousel required in a right way.

### DIFF
--- a/packages/carousel/src/components/app.tsx
+++ b/packages/carousel/src/components/app.tsx
@@ -76,6 +76,15 @@ const baseAuthoringProps = {
 
   fields: {
     iframeAuthoring: IframeAuthoring
+  },
+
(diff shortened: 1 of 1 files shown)


emcelroy (approved):
Looks good! I'm guessing it would either be difficult to disable the required checkbox on the subquestion, or there's some good reason not to that I'm not thinking of?

pjanik:
> Looks good! I'm guessing it would either be difficult to disable the required checkbox on the subquestion, or there's some good reason not to that I'm not thinking of?

We could probably uncheck these inputs programmatically. But I guess it could be a bit confusing. Also, I avoided that not to add some confusing code. I'd rather warn authors that they do something unexpected rather than try to programmatically fix their actions that are not supported.

PR #288, opened 2023-02-23
ci: Use 's3-deploy' Action & 'ubuntu-latest' (PT-184551778)

433eros:
PT-184551778, mostly copied from concord-consortium/activity-player#404

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,9 +87,8 @@ jobs:
           prefix: ${{ github.event.repository.name }} # 'question-interactives'
           awsAccessKeyId: ${{ secrets.AWS_ACCESS_KEY_ID }}
           awsSecretAccessKey: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-          # Add the default branch and the current branch as top branches
+          # Add the default branch as a top branch
           topBranches: |-
             [
-              "${{ github.event.repository.default_branch }}",
-              "${{ github.ref_name }}"
+              "${{ github.event.repository.default_branch }}"
             ]


pjanik (commented):
It looks good to me.

I could see https://models-resources.concord.org/question-interactives/branch/ubuntu-18.04-brownout/open-response/

However, somehow this returns 404:
https://models-resources.concord.org/question-interactives/branch/ubuntu-18.04-brownout/version.json

while master branch works fine:
https://models-resources.concord.org/question-interactives/branch/master/version.json

I only found it, as that was the URL I had in my browser history. :wink: Maybe it's worth checking manually the content of the deployed dir in models-resources bucket.

scytacki (changes requested):
I don't think this will correctly release the question interactives.
The `ci.yml` will correctly deploy the dist folder to the `version/[tag-name]`.
But the release.yml will try to copy some `version/[tag-name]/index-top.html` file which doesn't exist in question interactives. 

There are separate index files for each interactive. So you'll need something more like: https://github.com/concord-consortium/foss-simulations/blob/main/.github/workflows/release.yml
That approach is kind of crazy with a matrix build in order to handle releasing each of the sub folder index-top files, but it might be the best way to do it without making a new javascript action.

Also I'd guess the topBranches support in the s3-deploy-action also doesn't handle handle subfolders like this, but I'll need to take a look at its code, which I don't have time to do right now.

scytacki:
Yes it turns out that the s3-deploy-action supports topBranches for subfolders. It is documented here: https://github.com/concord-consortium/s3-deploy-action#top-branch-support-in-mono-repos

scytacki:
So if everything was working properly, this would be the place to test the "released" open-response:
https://models-resources.concord.org/question-interactives/open-response/index-ubuntu-18.04-brownout.html

But I can see that file doesn't exist:
https://s3.console.aws.amazon.com/s3/buckets/models-resources?region=us-east-1&prefix=question-interactives/open-response/&showversions=false

Which is because you didn't change the build system to create `index-top.html` files for each of the subfolders.

Generally we don't use the released question-interactives. Inside of LARA we point directly to the versioned path for each interactive so that would be: 
https://models-resources.concord.org/question-interactives/version/v1.11.0/open-response/index.html

However there is at least one place where the released version is used. This is the `full-screen` interactive. So it is important to create the index-top.html files, and update the release.yml to copy them. 

The full-screen interactive doesn't change often, so if this becomes urgent because of the brown out, we could merge this as is and fix that stuff afterwards.

scytacki:
@433eros yes it won't be there because the webpack build files have to be updated to generate `index-top.html` files. The Activity Player does that here: 
https://github.com/concord-consortium/activity-player/blob/master/webpack.config.js#L131-L135 

I don't remember the new question interactives build system well enough to tell you off hand how best to update it. Perhaps @pjanik can either do this or tell you what to do.

scytacki (changes requested):
This is looking better but the index-top files don't have the right paths to their assets.
If you load:
https://models-resources.concord.org/question-interactives/open-response/index-ubuntu-18.04-brownout.html
And look in the console, you'll see it is failing to load its css and javascript assets. they have an extra `open-response/` in them. 

You can also look at the source of the index-top.html that is generated and see it has assets URLs like:
`../branch/ubuntu-18.04-brownout/open-response/open-response/assets/index.d851ed95937a61183eae.css`

I believe the right fix is to remove the `interactiveName` out of the publicPath.

cc @pjanik

pjanik:
It's also missing report-item/index.html and version.json.
report-item/index-top.html should be added to webpack config. It's gonna look pretty much like a regular index-top.html config that you added.

Then, it needs to be copied in release.yml. Take a look how ZIP files are copied here:
https://github.com/concord-consortium/foss-simulations/blob/main/.github/workflows/release.yml#L50
You could reuse a similar pattern to copy report-item/index-top.html and version.json files.

.github/workflows/ci.yml
```diff
@@ -64,29 +64,32 @@ jobs:
...
+          bucket: models-resources
+          prefix: ${{ github.event.repository.name }} # 'question-interactives'
+          awsAccessKeyId: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          awsSecretAccessKey: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          # Add the default branch and the current branch as top branches
+          topBranches: |-
+            [
+              "${{ github.event.repository.default_branch }}",
+              "${{ github.ref_name }}"
+            ]
```
scytacki, line 95:
```suggestion
          # Add the default branch as a top branch
          topBranches: |-
            [
              "${{ github.event.repository.default_branch }}"
            ]
```
Before merging the current branch should be removed, so we don't end up with separate `index-[branch]`.html` files for each branch going forward.

PR #289, opened 2023-02-24
feat: Added DynamicText tags for read aloud (PT-184531998)

dougmartin:
Added DynamicText tags around all read aloud elements.


scytacki (approved):
When we looked at this last together it seemed good to me.
With the squashed commits it isn't easy for me to see what you've changed since the last time, but I trust you. :)

PR #290, opened 2023-03-02
fix: Fixed some of the commented out image-question runtime tests

dougmartin:
The tests that can be changed from shallow() to render() were fixed.  Three tests that require mounts() can't be fixed.  The reason is given in a comment above those tests.


scytacki (approved):
Seems fine to me.

PR #293, opened 2023-03-08
fix: Fixed issues found by QA for read aloud

dougmartin:
- [PT-184630913] Multiple Choice - Icon and feedback text are misaligned
- [PT-184640489] Bar values are not displayed above the slider
- [PT-184531998] Fixed read aloud of locked question text

Also upgraded to latest dynamic-text package to stay in sync with AP.  There is no new functionality used in the new version but its best to keep it synced to avoid confusion in the future.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3181,9 +3181,9 @@
       }
     },
     "node_modules/@concord-consortium/dynamic-text": {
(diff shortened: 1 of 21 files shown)


pjanik (approved):
Looks good 👍

PR #294, opened 2023-03-21
fix: Remove unneeded scrollbars from Labbook thumbnail chooser [PT-184687743]

dougmartin:
Hides the scrollbars in the Labbook thumbnail chooser as it already has arrow controls to scroll through the thumbnails.


pjanik (approved):
Looks good 👍

PR #295, opened 2023-03-22
feat: Added logging of labbook comments [PT-179239685]

dougmartin:
Added logging on labbook comments, trigger in two ways:

1. When the comment textarea is blurred and the comment has changed the change is logged.

2. When the comment changes the comment is logged after a 10 second delay (to avoid flooding the logs) unless it is has already been logged by the blur event.  This 10 second delay is reset on each change to the text area.

### DIFF
--- a/packages/labbook/src/components/comment-field.tsx
+++ b/packages/labbook/src/components/comment-field.tsx
@@ -1,7 +1,8 @@
-import React, { ChangeEvent } from "react";
+import React, { ChangeEvent, useCallback, useEffect, useState } from "react";
 import { ThumbnailTitle } from "./thumbnail-chooser/thumbnail-title";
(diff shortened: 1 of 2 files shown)


pjanik (approved):
Looks good. But isn't 10 seconds too long? A blur event might not happen, or there might not be enough time if the user is going to change pages or close the tab. I think some other components use 2 seconds delay (eg. AP answer saving). But I guess we never really discussed or tested what's the "right" value.

dougmartin:
@pjanik Good point.  I changed it to 2 seconds here in this follow on commit: https://github.com/concord-consortium/question-interactives/pull/295/commits/79f68c8ccbaa1c49c25850ae5f8df86ebaeae394

PR #297, opened 2023-04-11
feat: Converted all font sizes to use rems with pxToRem helper [PT-184910439]

dougmartin:
*NOTE*: there are a few css issues reported after initial testing but these will be addressed by a follow on PR.  Once this is merged the lara-interactive-api will be updated in the next PR to add the font size and once that is done the CSS fixes will be easier to complete (as opposed to having to rebase the existing spike PR that forced the font size).


scytacki (approved):
looks good to me

PR #300, opened 2023-04-27
fix: Authoring form styling after font size change to rems [PT-185043384]

dougmartin:
This is a sledgehammer approach to fixing the issue of the authoring form styling breaking after the bootstrap file was converted to scss and the fonts were changed to rems from px to accomodate the font-size feature.

This fix adds a second copy of the original unmodified bootstrap.css file within a :global section of the authoring.scss file.  This locallizes the original bootstrap layout to the authoring form area ensuring it lays out the same as it did before the font size update.

### DIFF
--- a/packages/helpers/src/styles/authoring.scss
+++ b/packages/helpers/src/styles/authoring.scss
@@ -1,4 +1,13 @@
 .authoring {
+  /*
+    This is a sledgehammer approach to fixing the issue of the authoring form styling breaking after the bootstrap
+    file was converted to scss and the fonts were changed to rems from px to accomodate the accessbility font-size
+    feature.
+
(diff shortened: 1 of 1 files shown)


scytacki (changes requested):
Seems like the best way to fix the problem in a short amount of time.

I'd just ask that a comment is added so when we find this in the future we have some clue about why it is there.

dougmartin:
@scytacki I've just added the comment in a follow on comment here: https://github.com/concord-consortium/question-interactives/pull/300/commits/fb28681d3c72f1e1c0dacd018fff3f91e15f633e

scytacki (approved):
Looks good.

PR #301, opened 2023-05-03
fix: Updated font-size issues found during QA [PT-184981533]

dougmartin:
- Changed scorebot title from px to rems
- Changed feedback container text from px to rems
- Changed audio recording icons to use px size to match height
- Changed upload button to not wrap text so it doesn't overlap the labbook


scytacki (approved):
seems fine to me.

PR #305, opened 2023-05-05
fix: Changed iframe runtime to accept readOnly property [PT-184981533]

dougmartin:
This fixes an issue with the font-size of scafolded questions after they are submitted as it was then showing the iframe in report mode which doesn't support the font-size change.

To set the iframe as read only pointer events are disabled and the tab index is removed to prevent tabbing into the iframe.

The carousel interactive was updated to use this new prop instead of adding a disabled div around the iframe.


scytacki (changes requested):
Is this "pointer events: none" approach close to what happens when the various question types are submitted outside of a scaffolded question?

Do they have some kind of "submitted" view that is different than report or this new readOnly approach?

dougmartin:
The open response uses a similar `readOnly` approach seen set here: 

https://github.com/concord-consortium/question-interactives/blob/master/packages/open-response/src/components/runtime.tsx#L24

and used to disable input here:

https://github.com/concord-consortium/question-interactives/blob/master/packages/open-response/src/components/runtime.tsx#L234-L245

I think the use of `report` view for the scaffolded question was just a way to implement cross iframe "readonlyness".  I think this approach is cleaner, and is needed for the font-size support since it only is supported in the runtime view.

PR #306, opened 2023-05-08
fix: added DynamicTextContext provider to demo [PT-185108429]

dougmartin:
Without the DynamicTextContext provider the demo would break for open response when the hint option was enabled due to its use of DynamicText.

### DIFF
--- a/packages/helpers/src/components/demo.tsx
+++ b/packages/helpers/src/components/demo.tsx
@@ -3,6 +3,7 @@ import { useAccessibility } from "@concord-consortium/lara-interactive-api";
 
 import { IframeRuntime } from "../components/iframe-runtime";
 import { DemoAuthoringComponent } from "./demo-authoring";
+import { DynamicTextContext, useDynamicTextProxy } from "@concord-consortium/dynamic-text";
 
 import css from "./demo.scss";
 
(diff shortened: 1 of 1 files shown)


packages/helpers/src/components/demo.tsx
```diff
@@ -17,6 +18,8 @@ interface IProps<IAuthoredState, IInteractiveState> {
 const iframe = new URLSearchParams(window.location.search).get("iframe");
 const rootDemo = !iframe;
 
+const dynamicTextProxy = useDynamicTextProxy();
```
scytacki, line 21: I'm surprised the linter isn't complaining about something that looks like a hook being used outside of a function. 🤷

scytacki (approved):
looks good

packages/helpers/src/components/demo.tsx
dougmartin: Its a poorly named function.  It's not a hook but a getter to return a singleton of a class:

https://github.com/concord-consortium/dynamic-text/blob/main/src/proxy.ts#L44

kswenson: But the `rules-of-hooks` linter rule assumes that a function that starts with `use...` _is_ a hook, which is the puzzling part of @scytacki's question. My suspicion would be that the rules of hooks are only applied within React components. 🤷

PR #307, opened 2023-05-08
fix: Voice typing authoring switch should not disappear when required is checked [PT-185100790]

dougmartin:


scytacki (approved):
loos good

PR #308, opened 2023-05-08
fix: Default answer disappearing in authoring [PT-185100790]

dougmartin:
The default answer was disappearing in open response authoring when the required checkbox was selected as it was not specified in the UI order AND the authoring iframe useAutoHeight() hook was not updating the height when the prediction feeback field showed after required was selected.

This change sets the UI order and makes the prediction feedback always show at the bottom of the form so it is not cut-off from view.

### DIFF
--- a/packages/open-response/src/components/app.tsx
+++ b/packages/open-response/src/components/app.tsx
@@ -39,35 +39,11 @@ const baseAuthoringProps = {
       defaultAnswer: {
         type: "string",
         title: "Default answer"
-      }
-    },
-    dependencies: {
-      required: {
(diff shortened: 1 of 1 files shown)


scytacki (changes requested):
Seems fine to me. 

There used to be some problem with authors not being able to see all of the authoring form when working with teacher edition. I'm not sure if that is still a problem, can you check how this change looks when the question is wrapped in a teacher edition?

dougmartin:
Merged after reply from @scytacki on Slack:

> Yeah it looks fine. :+1:

PR #310, opened 2023-05-09
feat: Added logging of voice typing [PT-185033795]

dougmartin:
This adds logging for starting and stoping the voice text feature along with updates as voice transcripts appear.

This also adds a console.log for the log messages in the generic demo container.

### DIFF
--- a/packages/open-response/src/components/runtime.tsx
+++ b/packages/open-response/src/components/runtime.tsx
@@ -65,6 +65,7 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
   const selectionRef = useRef<{start: number, end: number}>({start: 0, end: 0});
   const voiceTypingRef = useRef<VoiceTyping>();
   const initialTextRef = useRef("");
+  const lastTranscriptRef = useRef("");
 
   useEffect(() => {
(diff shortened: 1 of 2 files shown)


packages/open-response/src/components/runtime.tsx
```diff
@@ -110,13 +111,19 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
...
+        if (transcript !== lastTranscriptRef.current) {
+          const selection = selectionRef.current || {start: 0, end: 0};
+          const updatedText = [
+            initialTextRef.current.substring(0, selection.start),
+            transcript,
+            initialTextRef.current.substring(selection.end)
+          ].join("");
+          handleUpdateTextArea(updatedText);
+          lastTranscriptRef.current = transcript;
+          log("voice typing updated", {voiceText: transcript, updatedText});
```
scytacki, line 125: If I understand this right, it seems like it will be a lot of log message each time the text area is updated. That seems like a lot of log messages to record. I'm pretty sure when the user is typing we buffer this a bit and don't record every change to the text area.

But if this is what Dan needs, then it is fine.

scytacki (changes requested):
Maybe too many log messages?

PR #313, opened 2023-05-15
fix: Removed updatedText field in voice typing log [PT-185033795]

dougmartin:
This will both reduce the amount of data logged and it is also redundant as the final text is logged when the voice typing button is toggled off.


scytacki (approved):
seems fine

PR #314, opened 2023-05-15
fix: Text/button overlapping image question dialog textarea [PT-185160708]

dougmartin:
This change removes the constraint for the dialog height and moves the close section into the right panel allowing the absolute positioning that was causing the overlap to be removed.

BEFORE:

image

AFTER:

image


dougmartin:
@scytacki I've added before and after screenshots to the PR.

scytacki (approved):
Looks good

PR #315, opened 2023-05-15
WIP: feat: Updated UI for audio recording [PT-185083392]

dougmartin:
WIP: the svg icons need to be re-exported so they can be targeted with hover fills.


lublagg (approved):
Looks good 👍

PR #317, opened 2023-05-31
feat: Added Lab book demo [PT-185302091]

dougmartin:
This will allow for faster development and demoing features.


lublagg (approved):
Looks good!

PR #318, opened 2023-05-31
fix: Set minium value for Lab book authoring number inputs [PT-184821574]

dougmartin:
This sets the minium value of the "Maximum number of lab book entries" and "How many thumbnails to display" number inputs to 1.  Previously there was no minimum which allowed for zero or negative values.


lublagg (approved):
Looks good!

PR #319, opened 2023-06-01
feat: Add start of new Lab Book layout for notebook [PT-185267555]

dougmartin:
This adds a optional query string parameter of "layout" that when set to "notebook" changes the layout.

Further layout changes may be made in future commits.


lublagg (approved):
Looks good

PR #321, opened 2023-06-05
fix: Changed play/pause open response audio logging [PT-185083597]

dougmartin:
- The pause is now logged instead of "stopped" and the play ended function, which logged a further event, is no longer called.  This was a bug introduced when the stop button handler was reused for pause.
- The play and pause logging now add the current audio timestamp for more context.


lublagg (approved):
Looks good 👍

PR #322, opened 2023-06-05
Wide layout labbook styling (PT-185267555)

lublagg:


packages/labbook/src/components/runtime.tsx
```diff
@@ -231,9 +237,9 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
       <div><DynamicText>{renderHTML(authoredState.prompt)}</DynamicText></div>
     }
     <div className={css["app"]} ref={containerRef}>
-      <div className={css["container"]} style={{width: containerWidth}}>
+      <div className={`${css["container"]} ${isWideLayout && css["wide"]}`} style={{width: containerWidth}}>
```
dougmartin, line 240: Same comment as above to switch to the `classnames` package.

packages/labbook/src/components/thumbnail-chooser/thumbnail.tsx
```diff
@@ -10,12 +10,14 @@ export interface IThumbnailProps {
   thumbContent?: React.ReactNode;
   data: any;
   onClick?: ()=> void;
+  wideThumbnail?: boolean;
 }
 
 export const Thumbnail: React.FC<IThumbnailProps> = (props:IThumbnailProps) => {
-  const {thumbContent} = props;
+  const {thumbContent, wideThumbnail} = props;
+  console.log("wideThumbnail", wideThumbnail);
```
dougmartin, line 18: Did you mean to leave this `console.log` in?

dougmartin (changes requested):
Looks good overall :+1:  You just need to use the `classnames` package to generate the class names instead of using string interpolation.

PR #324, opened 2023-06-09
fix: update LARA image proxy URL to use production [PT-185233166]

pjanik:
When I was testing it myself, I experienced https://lara-master.concordqa.org returning 503 error (both the main index page and image-proxy), and a moment later everything was fine. So I think that's the main reason why backgrounds don't work in a consistent way.


scytacki (approved):
Looks good. 👍

PR #327, opened 2023-06-12
fix(DrawingTool): better support of staging Token Service environments

pjanik:
[PT-185233166]

This PR will allow us to reconfigure Token Service to use `https://token-service-files.concordqa.org/` domain instead of `https://token-service-files.s3.amazonaws.com/`. 

Note that there's another bug somewhere around that makes Token Service **always** use the staging / QA environment - even on LARA production. But this will improve things on staging, and it'll slightly improve things on production until the other bug is fixed.


scytacki (approved):
This looks good to me.

pjanik:
@scytacki, I just removed the direct token-service-files S3 bucket URL from the proxy whitelist as we discussed. It started to work for me, but just to be safer. It's not a regression, as we've been always using proxy in this case. I was just hoping we can optimize it.

scytacki (approved):
Looks good

PR #329, opened 2023-06-13
feat: Add "export to media library" setting [PT-184813635]

dougmartin:
This uses a common helper to generate the authoring schema settings.


lublagg (approved):
Looks good 👍

PR #331, opened 2023-06-14
fix: Removed times from audio logs [PT-185083597]

dougmartin:
The extra time parameter that was added to the audio logging during the audio UI change was not requested and is buggy so it is being removed.


lublagg (approved):
Looks good 👍

PR #333, opened 2023-06-14
Fix input text styling (PT-185345611).

lublagg:
This PR makes it so that in Image Question, only placeholder text in a textarea is italicized. The student answer typed in and displayed after are not italicized.


pjanik (approved):
Looks good 👍

PR #335, opened 2023-06-15
feat: Added font type from accessibility [PT-185336244]

dougmartin:
This updates the options to the useAccessibility() hook to target the runtime div for dynamic font-family update.

It also removes a few unnecessary explicit Lato font-family references so that the font-family can be updated in one place for the notebook.


lublagg (approved):
Looks good 👍

PR #341, opened 2023-06-23
fix: Drawing tool in image question getting clipped on iPads [PT-185414442]

dougmartin:
This resizes the image question drawing tool elements when displayed on a iPad sized window.


lublagg (approved):
Looks good 👍

PR #342, opened 2023-06-27
Students can select images from Activity Media Library (PT-184813641)

lublagg:
This PR makes changes to the Labbook, Image Question, and Drawing Tool interactives. 

If an author allows a user to upload an image from the Activity Media Library for a particular interactive, instead of the regular upload pattern, a user is presented with a dialog that gives them the option to either upload from their own device or to pick an image from the activity media library.

Because some of the components are used across all three interactives, some of them are located in the helpers folder.


dougmartin (approved):
Looks great :+1:

PR #343, opened 2023-06-30
Upgrade lara-interactive-api to ^1.9.3.-pre.8 (PT-185453872)

lublagg:
This upgrade is necessary because in 1.9.3.-pre.1 + above of lara-interactive-api, iframe-phone was being imported from the incorrect place. This was causing interactive state saving to be cancelled when the drawing tool dialog was opened after trying to upload an image or take a snapshot in image question.

### DIFF
--- a/packages/bar-graph/package.json
+++ b/packages/bar-graph/package.json
@@ -107,7 +107,7 @@
   },
   "dependencies": {
     "@concord-consortium/dynamic-text": "^1.0.2",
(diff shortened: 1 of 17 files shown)


pjanik (approved):
Looks good. 

Semi-related, but I'm wondering if we shouldn't finally release a regular v1.9.3 of lara-interactive-api. :wink:

PR #344, opened 2023-07-03
fix: Update multiple choice to use lara-interactive-api 1.9.3.pre.8 [PT-185302126]

dougmartin:
The dependency was out out of date in this package.


lublagg (approved):
Good catch :)

PR #345, opened 2023-07-04
Image media library QA fixes (PT-184813641)

lublagg:
This PR addresses the following issues:

-  Labbook: If media library selection option is disabled and if student reached the maximum number of entries, Replace Current Image and Create New Image buttons are disabled. Unable to proceed further. (User should be able to click on the Replace Current Image button)
- Labbook: If media library selection is enabled and if student reached the maximum number of entries, instead of Replace Current Image and Create New Image buttons, getting Please Wait button in disabled state. Unable to proceed further. (User should be able to click on the Replace Current Image button)
- In image question & drawing tool, if question is set to required, able to submit the answer when dialog is displayed. (User should not be able to click the submit button when the dialog is open)
- In Image question & drawing tool interactives, when secondary column is visible, in 60-40/40-60 layouts, dialog is getting cut off on the right & in 70-30/30-70 layouts, dialog is not center aligned. (Dialog should always be visible)
- When font size set to large, alignment issue occurs in the dialog and Upload an image from this activity: heading not supports the large font settings. 

I also compartmentalized the image-question runtime into a few different components for readability's sake.

### DIFF
--- a/packages/drawing-tool/src/components/runtime.tsx
+++ b/packages/drawing-tool/src/components/runtime.tsx
@@ -80,14 +80,18 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
         <TakeSnapshot authoredState={authoredState} interactiveState={interactiveState} setInteractiveState={setInteractiveState} />
       }
       { showUploadModal &&
-        <UploadFromMediaLibraryDialog
-          onUploadImage={handleUploadImage}
-          onUploadStart={handleUploadStart}
-          mediaLibraryItems={mediaLibraryItems}
-          uploadInProgress={uploadInProgress}
-          setUploadInProgress={setUploadInProgress}
-          handleCloseModal={() => setShowUploadModal(false)}
-        />
+        <>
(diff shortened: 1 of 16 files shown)


packages/drawing-tool/src/components/runtime.tsx
```diff
@@ -80,14 +80,18 @@ export const Runtime: React.FC<IProps> = ({ authoredState, interactiveState, set
...
-          onUploadImage={handleUploadImage}
-          onUploadStart={handleUploadStart}
-          mediaLibraryItems={mediaLibraryItems}
-          uploadInProgress={uploadInProgress}
-          setUploadInProgress={setUploadInProgress}
-          handleCloseModal={() => setShowUploadModal(false)}
-        />
+        <>
+          {/*create an empty div with needed height so modal can overlay properly*/}
+          <div style={{height: "600px"}}/>
```
pjanik, line 85: Some CSS wouldn't work here?

pjanik (approved):
Looks good 👍

PR #346, opened 2023-07-04
Fix Image Question state saving issues

pjanik:
https://www.pivotaltracker.com/story/show/185453872

I spent quite a lot of time debugging this issue and yes, it was extremely confusing.

I think that something went pretty bad with lockfile and workspace dependencies, maybe when there was a switch from `lara-interactive-api` 1.9.2 to the next version. I've noticed that `lara-interactive-api` was installed in two places:
- top-level `node_modules`, version **1.9.2**
- inside every question type directory (eg. `packages/image-question`), version **1.9.3-pre.X**

Most likely this means that the library was imported twice. 
And if you check `showModal` call, it relies on `flushStateUpdates` that ensure that the interactive state is updated **before** modal is shown. I've noticed that it was not happening. My bet is that two copies of the same library were messing up the module-level reference to state update that should be flushed.

Once I totally removed all the node_modules dir and package-lock.json, then ran `npm i` and let it generate the new lock-file, the `lara-interactive-api` was installed only in one place - top-level node_modules. This also fixed the image question issue.

But... it introduced a confusing issue with the rich text editor widget in Cypress tests. Cursor jumps in the Cypress tests. I was looking if I can fix it somehow, but new version of slate editor or Cypres was not helping. So, I've found a not-so-great workaround with adding a white-space at the beginning of the string. I bet that a proper fix of the issue would require some work in slate-editor. But it seems to affect only test env / Cypress.


pjanik:
There's not that much to review here in fact - I'm gonna merge this PR, so it's easier for Tejal or Saravanan to smoke test all the interactives in the master branch (there are already library interactives pointing to master). If you have some comments, please leave them even after it's merged and I'll address them. :)

PR #347, opened 2023-07-05
Add "answerPrompt" class to drawing-tool-dialog css.

lublagg:
Add this css class to the drawing-tool-dialog css file, so the AP cypress test can find the element.


pjanik (approved):
Looks good 👍 
But if it's just for tests, are the margins also necessary?

Generally, we could also switch to `data-testid` attributes to make it less fragile.

lublagg:
@pjanik The margins seem to match the specs, so I think it's OK.

PR #349, opened 2023-07-10
build: Updated lara-interactive-api to 1.9.4 [PT-185396456]

dougmartin:


pjanik:
Looks good 👍 
It's definitely nice to have, as `lara-interactve-api` 1.9.4 added a check that tests if only one copy of `lara-interactve-api` is imported. But as I mentioned on Slack - no bug fixes, features, or PT stories depend on that. So, I don't think we need to release QI v1.15.1 or anything like that.

pjanik (approved):

PR #350, opened 2023-07-10
fix(IframeAuthoring): handle case when formData is undefined [PT-185560197]

pjanik:
I've verified manually that the previous version of reactjson-schema-form was providing `formData={}` while now it's `formData=undefined`. I don't know where this change originated, but this PR should maintain the old behavior.


dougmartin (approved):
Looks good :+1:

PR #351, opened 2023-07-26
fix: Labbook comment text layout in reports [PT-185471385]

dougmartin:
This fixes an issue in the portal reports where long comments did not wrap when displayed.  This also makes the comments 100% of the width in the reports and adds more padding as requested.


pjanik (approved):
Looks good 👍

PR #352, opened 2023-07-26
feat: Add authoring option to disable the Labbook annotation tool [PT-185699750]

dougmartin:
The new authoring option defaults to false.  When set to true the annotation tool is not shown in the labbook toolbar.

The new authoring option looks like this, as requested:

image


pjanik (approved):
It looks good, but I'm surprised you had to add a custom widget to make a checkbox. There's nothing provided by react jsonschema forms?

PR #353, opened 2023-10-23
fix: In iPad Lab book thumbnails have the close button icon missing [PT-185414388]

dougmartin:
- Adds missing `padding: 0`, which is needed for iPad buttons with svg children
- Nudges the X svg child element over 1 pixel on iPads, otherwise the centering is a bit off


eireland (approved):
👍 LGTM

PR #354, opened 2023-10-25
fix: Labbook: Upload image modal dialog should match the spec [PT-185462804]

dougmartin:
- Updated the button padding to match the spec
- Fixed missing import of px-to-rem
- Added label to new thumbnail


lublagg (approved):
LGTM 👍

PR #359, opened 2023-11-01
186280538 snapshot dialog bug fixes

eireland:
Bug fixes as described in this[ PT Story](https://www.pivotaltracker.com/story/show/186280538). Note that only #1, 3, 4, 5 & 6 are addressed here. 
- Removes snopshot icon in dialog
- Text when max images is reached now reads `"Create Snapshot is not available..."
- Creat Snapshot button is disabled if max images is reached
- Button text is centered.

Not described in story but is fixed:
- When font is large, modal height is adjusted so that all content fits in the modal.


lublagg (approved):
Nice - looks good to me 👍

PR #360, opened 2023-11-03
fix: Labbook media library button styling [PT-185462804]

dougmartin:
This centers the labbook media library upload buttons.


lublagg (approved):
Looks good 👍

PR #361, opened 2023-11-04
186280538 fix snapshot wording

eireland:
Fixes wording when can't create a new image
Fixes button wrapping when in large font, and in notebook mode.


packages/labbook/src/components/upload-button.tsx
```diff
@@ -7,12 +7,11 @@ export interface IUploadButtonProps {
...
   disabled?: boolean;
   onClick?: (type: string) => void;
-  inDialog?: boolean;
 }
 
 export const UploadButton: React.FC<IUploadButtonProps> = (props) => {
-  const {children, onClick, disabled, inDialog} = props;
-  const classes = classNames(css["upload-button"], {[css.disabled]: disabled, [css["in-dialog"]]: inDialog});
+  const {children, onClick, disabled} = props;
+  const classes = classNames(css["upload-button"], {[css.disabled]: disabled});
```
dougmartin, line 14: I think this may break the final fix I made for media upload buttons that QA found?  Could you test this with an activity that exports an image to the media library and then enables the media library in the labbook?  You'll need to first select a media library image and then replace it and check the button padding/size.

dougmartin (commented):
Before I approve could you check the comment I added about testing this with media library items?

packages/labbook/src/components/upload-button.tsx
eireland: Added back the inDialog css. Don't even remember why I removed it.

dougmartin (approved):
Looks good :+1:

PR #364, opened 2024-01-08
feat: Authors can customize image question drawing tools [PT-182016906]

dougmartin:
This adds a list of checkboxes that can be used to hide drawing tools.

These checkboxes are showing in the drawing tool, the lab book and the image question.

The hide annotation tool option in the labbook has been removed and added to this new control.

This also adds a demo to the drawing tool for ease of development.

Screenshot:
---

image


lublagg (approved):
Looks good to me 👍

PR #365, opened 2024-01-09
Moved placement of hide drawing tools authoring widget to always be just above the media library options

dougmartin:


lublagg (approved):
👍

PR #366, opened 2024-01-09
fix: delete of hideAnnotationTool in labbook authoredState

dougmartin:
Adds conditional to ensure the property exists before it is deleted to prevent an error from being thrown.


lublagg (approved):
👍

PR #369, opened 2024-01-16
fix: Hide toolbar QA issues [PT-182016906]

dougmartin:
- Changed : to . in hide tools option form hint
- Changed "Stroke Width" to "Stroke width" in hide tools options


lublagg (approved):
Looks good :)

PR #375, opened 2024-05-03
182721195 save side-by-side interactive state on nav

eireland:


pjanik (commented):
Generally looks good, especially if it fixes the bug. :)

Out of curiosity, do we understand why the ref had to be used there? What was happening without it? 

I'm asking because we're touching a pretty critical part of the question interaction and answer saving. It would be nice to understand that completely. Otherwise, we might accidentally break the saving of answers in some edge cases. Perhaps @dougmartin should double-check that, as I guess Doug is more familiar with the whole onUnloadCallbackRef and resolveOnUnload logic.

packages/helpers/src/components/iframe-runtime.tsx
```diff
@@ -74,17 +76,19 @@ export const IframeRuntime: React.FC<IProps> =
...
+
       phone.addListener("interactiveState", (newInteractiveState: any) => {
         setInteractiveStateRef.current?.(newInteractiveState);
         if (flushOnSave) {
           // don't wait the default 2000ms timeout before saving
           flushStateUpdates();
         }
-        if (onUnloadCallback && resolveOnUnload.current) {
+
+        if (onUnloadCallbackRef?.current && resolveOnUnload.current) {
```
dougmartin, line 87: I think you can change this to `onUnloadCallbackRef.current` as the current value is always set on line 56 (the current value may be undefined but the ref is always defined).

packages/helpers/src/components/iframe-runtime.tsx
```diff
@@ -74,17 +76,19 @@ export const IframeRuntime: React.FC<IProps> =
...
           // send the interactive state to any parent interactive that has provided an onUnload
           // callback, and then resolve the promise that was saved in the setOnUnload function
           // with undefined so the parent interactive doesn't have its state overwritten.
-          onUnloadCallback(newInteractiveState);
+          onUnloadCallbackRef?.current(newInteractiveState);
```
dougmartin, line 91: Same comment as above, should be `onUnloadCallbackRef.current`.

dougmartin (changes requested):
Looks good overall :+1:  I'm really surprised we didn't catch the `onUnloadCallback` being bound inside the interactive state listener.  We need to smoke test this change with all the interactives in QA to ensure this doesn't break them.

Could you change the check for the `onUnloadCallbackRef` as mentioned in the comments?  I think it will make this easier to understand for someone in the future.

dougmartin (approved):
Changes look good :+1:

PR #380, opened 2024-06-25
build: Update drawing-tool lib to 2.3.1 [PT-187848326]

dougmartin:
This updates the drawing-tool library to 2.3.1 which fixes crashes with annotations (outlined in PT-185363141).


pjanik (approved):
Looks good 👍

PR #384, opened 2024-07-11
ImageQuestion: fix Cancel button behavior

pjanik:
https://www.pivotaltracker.com/story/show/187941022

Setting `userBackgroundImageUrl: undefined` made sense when the dialog was opened right after upload, but it didn't make much sense in other cases, e.g., when the dialog was opened multiple times and "Cancel" was not related to the upload action.


scytacki (approved):
Seems fine to me.

PR #385, opened 2024-07-11
Image Question: do not autosave Drawing Tool content, so Cancel button can work as expected

pjanik:
https://www.pivotaltracker.com/n/projects/2441249/stories/185403513

@dougmartin, I'm marking you as a reviewer because you apparently worked on the same story last year. I was fixing other things when I randomly noticed this issue, and since I was in the right context, it seemed possible for me to fix it.

Can you check it against your comments from the PT story, if you remember anything about that? I don't think I touched the code that is shared with Labbook, as I believe the `DrawingToolDialog` component is used only by the Image Question.

### DIFF
--- a/packages/image-question/src/components/drawing-tool-dialog.tsx
+++ b/packages/image-question/src/components/drawing-tool-dialog.tsx
@@ -12,7 +12,8 @@ import css from "./drawing-tool-dialog.scss";
 
 interface IProps {
   authoredState: IAuthoredState;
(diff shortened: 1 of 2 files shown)


dougmartin (approved):
Looks good :+1:

dougmartin:
@pjanik I looked at the PT story but can't really remember the context I made those comments.  Sorry.

PR #386, opened 2024-10-04
fix: Image question answers with no answer text save failing [PT-188258961]

dougmartin:
On questions with no previous answers the previous state is undefined with cases the setInteractiveState handler to fail.  The fix is to add a optional chaining operator.


emcelroy (approved):
Looks good 👍

PR #388, opened 2024-10-16
fix: Hide submit button when draw tool dialog showing [PT-188398397]

dougmartin:
The submit button, rendered in the BaseQuestionApp component, was showing at the bottom of the draw tool dialog in the image question app.

This fix centralizes the url param code that is used to construct the url show in the drawing tool dialog and then sets the existing (optional) disableSubmitBtnRendering prop to the BaseQuestionApp to hide the submit button when the current url has the drawingToolDialog url param present.

This can be tested using the built branch with this activity that has both the branch and the master branch image questions:

https://activity-player.concord.org/branch/master/?activity=https%3A%2F%2Fauthoring.lara.staging.concord.org%2Fapi%2Fv1%2Factivities%2F1187.json&author-preview=true


bgoldowsky (approved):
LGTM 👍

PR #390, opened 2025-04-30
feat: add deployment link to Jira (START-1)

emcelroy:
[START-1](https://concord-consortium.atlassian.net/browse/START-1)

Add `githubToken` and `deployRunUrl` to `.github/workflows/ci.yml`. This will allow the new version of `s3-deploy-action` to create a GitHub Deployment and set its `log_url` property, which will allow Jira to add a deployment link to associated Jira issues.

[START-1]: https://concord-consortium.atlassian.net/browse/START-1?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,6 +87,8 @@ jobs:
           prefix: ${{ github.event.repository.name }} # 'question-interactives'
           awsAccessKeyId: ${{ secrets.AWS_ACCESS_KEY_ID }}
           awsSecretAccessKey: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          deployRunUrl: https://models-resources.concord.org/question-interactives/__deployPath__/
           # Add the default branch as a top branch
           topBranches: |-
             [


.github/workflows/ci.yml
```diff
@@ -87,6 +87,8 @@ jobs:
           prefix: ${{ github.event.repository.name }} # 'question-interactives'
           awsAccessKeyId: ${{ secrets.AWS_ACCESS_KEY_ID }}
           awsSecretAccessKey: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          deployRunUrl: https://models-resources.concord.org/question-interactives/__deployPath__/
```
emcelroy, line 91: The resulting deployment link will return a 404 since there isn't an index page or anything at the root directory. I'm not sure what else to set this to, though. We could point it at one of the basic question interactive types, like Open Response, but I'm not sure that's a good idea.

scytacki: Here are a couple of long shots:

Perhaps the activity player has an option to set the base URL of the question interactives? 
Then you could point it at some demo activity that includes most of the question interactives.
So then the deployRunUrl would be something like `https://activity-player.concord.org/branch/master/?questionInteractveBase=https://models-resources.concord.org/question-interactives/__deployPath__/`

Perhaps the activity player supports loading interactives relative to the json file describing the activity. Then there could be an activity json located in the question-interactives repository which is deployed too. So then the deployRunUrl would be something like `https://activity-player.concord.org/branch/master/?activityUrl=https://models-resources.concord.org/question-interactives/__deployPath__/example-activity.json`

I doubt either option exist. In which case the 404 link seems fine to me.

scytacki (approved):
Looks fine to me. I left a comment about some long shot improvements, but I doubt they exist.

PR #393, opened 2025-05-13
fix: Remove trailing comma in scorebot feedback report [QI-13]

dougmartin:
This fixes a trailing comma that has been in the code that generates the scorebot feedback report for a couple of years but just now noticed.


emcelroy (approved):
👍

PR #397, opened 2025-09-17
feat: Add blockly interactive [QI-31]

dougmartin:
This commit adds a new Blockly interactive to the question interactives package.


tealefristoe (commented):
I think it would be good to merge this as is, and Ethan and I can clean up any loose ends in future PRs. However, we should probably fix the jest tests before doing that.

PR #503, opened 2026-05-20
Add tick-based sample interval to agent-simulation [QI-167]

dougmartin:
## Summary

QI-165 added a wall-clock sample-interval throttle to agent-simulation
recordings. QI-167 adds a tick-based option so authors can record one row
every N simulation ticks instead of every N milliseconds. Tick-based
throttling is useful when the sim's internal time (tick number) is the
meaningful unit rather than wall-clock time, for example when sim speed
varies or when comparing recordings from runs at different speeds.

## Authoring changes

The single `sampleIntervalMs` field is replaced by two fields:

- **Throttle samples by** (`sampleIntervalUnit`): a dropdown with "No
  throttling", "Every N milliseconds", "Every N simulation ticks".
- **Interval** (`sampleInterval`): a single integer value whose meaning
  depends on the dropdown.

`preprocessFormData` strips the interval value when the unit is "No
throttling". `maxSamples` (the per-recording cap) is unchanged.

## Runtime changes

The `afterTick` callback gains a tick-throttling branch alongside the
existing wall-clock branch. In tick mode it keeps every Nth sim tick
(`0, N, 2N, ...`); the first tick after a sim reset is always sampled.
Both throttle units compose with the `maxSamples` and `maxRecordingTime`
caps.

## Backward compatibility

The legacy `sampleIntervalMs` field is removed from the schema and from
`IAuthoredState`. A new `migrateAuthoredState` (V1 to V2) converts existing
activities on every read, following the carousel/labbook/scaffolded-question
pattern. The content-schema `version` bumps from 1 to 2. The migration is
read-only: the persisted JSON stays in legacy shape until the author next
saves.

## Testing

- 212 unit tests pass, including new tests for tick-mode sampling, the
  first-tick rule, tick-mode `maxSamples` auto-stop, free-play tick
  throttling, mid-run unit changes, the runtime no-op fallback, the
  V1 to V2 migration, and `preprocessFormData`.
- Lint and TypeScript checks pass for the agent-simulation package.
- Verified live in the demo harness across all throttle-unit and cap
  combinations (tick and ms throttle, each with `maxSamples` and
  `maxRecordingTime`).

## Spec

See `specs/QI-167-agent-sim-tick-based-sample-interval.md` for the closed
spec with full requirements and design decisions.

Jira: https://concord-consortium.atlassian.net/browse/QI-167


emcelroy (approved):
Looks good 👍

PR #511, opened 2026-06-29
feat: add accessible title to carousel child interactive iframes (QI-176)

emcelroy:
Child interactive iframes inside a Carousel question interactive were rendered without a `title` attribute, which is an accessibility gap (screen reader users rely on iframe titles to identify embedded content).

Unlike top-level interactives — which LARA / Activity Player give a `title` from the authored **Name** field — a Carousel's child interactives are rendered *inside* the Carousel itself and have no Name field in authoring. The hosts have no knowledge of these children, so the title has to be supplied by the Carousel.

## Changes

- **`IframeRuntime` (helpers) now always renders a `title`.** Added an optional `title` prop; when omitted it falls back to the embedded interactive's type name (`getLibraryInteractive(url)?.name`) and finally to `"Interactive content"`. Because this is the shared component used by all nested-interactive hosts, **side-by-side and scaffolded-question child iframes also gain a sensible title for free.**
- **Carousel supplies a position-aware title per slide** of the form `Slide N: <interactive type>` (e.g. `Slide 1: Open response`, `Slide 2: Image`), falling back to `Slide N: Interactive content` if the type can't be resolved.

No changes were needed in LARA or Activity Player — both only ever set the title on the top-level Carousel iframe and have no knowledge of its children.

## Why this approach

- Derived at render time from data already present in authored state → **works on all existing Carousels with no migration and no authoring changes.**
- Centralizing the fallback in `IframeRuntime` means every child iframe it renders gets a title, not just the Carousel's.

## Known limitation

Slides of the same type produce titles distinguished only by position (`Slide 1: Image`, `Slide 4: Image`). This is inherent to "type + position." If richer titles are wanted later, the per-slide derivation is the only place that would need to change (e.g. pulling the child's prompt text).

### DIFF
--- a/packages/carousel/src/components/runtime.test.tsx
+++ b/packages/carousel/src/components/runtime.test.tsx
@@ -1,0 +1,65 @@
+import React from "react";
+import { render, cleanup } from "@testing-library/react";
+
+import { Runtime } from "./runtime";
+import { IAuthoredState } from "./types";
+
+// Capture the props each child IframeRuntime is rendered with, so we can assert the title
+// the carousel derives for each slide.
+jest.mock("@concord-consortium/question-interactives-helpers/src/components/iframe-runtime", () => ({
+  IframeRuntime: (props: any) => <div data-testid="iframe-runtime" data-title={props.title} />,
+}));
+
+jest.mock("@concord-consortium/lara-interactive-api", () => ({
+  useAccessibility: jest.fn(() => ({})),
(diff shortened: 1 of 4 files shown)


kswenson (approved):
👍 Looks good — clean, focused, well-tested accessibility fix.

The change is purely additive (the child iframes previously had no `title` at all, so nothing existing is overridden), and centralizing the fallback in `IframeRuntime` means side-by-side and scaffolded-question children get a sensible title for free as a bonus.

PR #517, opened 2026-07-15
fix: trap keyboard focus in confirm-mode modals [QI-158]

kswenson:
## What & why

[QI-158](https://concord-consortium.atlassian.net/browse/QI-158) — trap keyboard focus in the Blockly file dialogs (Questions `#14` & `#15`). Audit issue `#95`, WCAG **2.4.3** (Level A, High).

Root cause was not in Blockly: the shared `Modal` (`packages/helpers`) gated **all** focus management behind alert mode, so confirm-mode dialogs got `role="dialog"`/`aria-modal` but no initial focus, focus trap, Escape handling, or focus restore. Fixing it in one place repairs every confirm-mode modal — the **five Blockly file dialogs** (New/Open/Copy/Rename/Delete) and agent-simulation's **Delete Recording**.

## Changes

**1. Shared `Modal` — extend focus management to confirm mode** (`modal.tsx`)
- Initial focus moves to the first focusable in the modal body (the text field, or **Cancel** when there's none — the safe default for a destructive confirm), skipping the title-bar Close (X). Alert mode still lands on OK.
- Escape routes to `onCancel` in confirm mode (dismiss without acting); alert mode keeps Escape → `onConfirm`.
- Tab trap and focus restore are now shared across both modes; the focusable selector is one constant.

**2. Blockly header — restore focus to the File menu button** (`header.tsx`)
- The file dialogs open from a menu item that unmounts when the menu closes, so the modal would capture a gone element and drop focus to `<body>` on close. Move focus to the persistent **File** button when a menu item is activated, so the dialog restores focus there on close (satisfies "focus returns to the trigger").

## Testing

- **Unit (TDD, RED→GREEN):** 5 new confirm-mode focus tests in `modal.test.tsx` (alert mode unchanged); new `header.test.tsx` for the focus-return. helpers 120/120, `maybe-file-modal` 46/46, agent-simulation 212/212, lint clean.
- **Real browser (Playwright/Chromium):** drove the New and Delete dialogs in the live interactive — initial focus correct, focus trapped across 12 Tab/Shift-Tab presses (never reaches the Blockly canvas), Escape dismisses, and focus restored to the File menu button. 10/10 checks.

Pre-existing Blockly `fetch is not defined` jsdom suite failures are unrelated (confirmed on `master`) and untouched by this PR.

## Out of scope

The broader File-menu keyboard operability — opening the menu via keyboard (it toggles on mouse-down only) and Enter/Space activation of items (WCAG 2.1.1) — is tracked in **QI-157**. The focus-return added here is a natural subset that ticket will build on.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[QI-158]: https://concord-consortium.atlassian.net/browse/QI-158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/blockly/src/components/header.test.tsx
+++ b/packages/blockly/src/components/header.test.tsx
@@ -1,0 +1,35 @@
+import React from "react";
+import { render, screen, fireEvent, cleanup } from "@testing-library/react";
+
+import { Header } from "./header";
+
+afterEach(cleanup);
+
+const defaultProps = {
+  name: "Model 1",
+  savedStates: [],
+};
+
+// The File menu button toggles the menu on mousedown (see header.tsx).
+const openMenu = () => fireEvent.mouseDown(screen.getByRole("button", { name: "File menu" }));
+
+describe("Header", () => {
+  it("returns focus to the File menu button when a menu item is activated", () => {
+    render(<Header {...defaultProps} onShowFileModal={jest.fn()} />);
+    const fileButton = screen.getByRole("button", { name: "File menu" });
+    openMenu();
+    fireEvent.click(screen.getByRole("menuitem", { name: /New/ }));
+    // The activating menu item unmounts with the menu, so focus must be moved to
+    // the persistent File menu button. That lets the dialog the parent renders in
+    // response capture it and restore focus there when the dialog closes.
+    expect(document.activeElement).toBe(fileButton);
+  });
+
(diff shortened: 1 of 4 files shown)


emcelroy (approved):
Looks good 👍

PR #522, opened 2026-07-22
feat: remove the unused button report-item scaffolding [QI-185]

dougmartin:
## What

Removes the dead report-item code path from the button ("I'm Done") interactive.

## Why

The button deliberately persists **no** interactive state: the single source of truth for a click's outcome is the report-service job record. A report-item can only render interactive state, so the button's report-item had nothing per-student to show. The files were scaffolded from the starter template but never given a real implementation (`get-report-item-html.tsx` returned `TODO` placeholder markup, and `report-item.test.tsx` was a stub with its body commented out), and nothing (no portal library record, curriculum manifest) ever registered `button/report-item` as a live report item.

## Changes

- Delete `packages/button/src/components/report-item/` (the micro-app, the `getReportItemAnswer` handler, the `TODO` HTML generator, and their two test files) plus the `report-item-index.tsx` webpack entry.
- Remove the report-item wiring from `webpack.config.js` (the `report-item` entry, its `HtmlWebpackPlugin`, and the `DEPLOY_PATH`-gated `report-item/index-top.html` plugin), collapsing to the standard two-entry shape.
- Drop the `getReportItemHtml` import and prop from `demo.tsx` (the demo no longer renders a report-item preview panel).
- Remove the now-unused `semver` dependency and `@types/semver` devDependency from `packages/button/package.json`, and regenerate the root `package-lock.json`.

No interactive-state persistence is added; the server-side job record stays the single source of truth. No behavior change for students or teachers.

## Verification (Node 18)

- `npm test` passes (3 suites, 43 tests)
- `npm run build` succeeds; `dist/button/` emits `index.html` + `demo.html` and **no** `report-item/`
- `npm run lint` clean
- `grep -rn "report-item" packages/button/src webpack.config.js` returns nothing

## Deploy notes

- No `.github/workflows/release.yml` change needed: its `report-item` promotion step is `test -f`-guarded, so it no-ops once the build stops emitting `report-item/index-top.html`.
- `package-lock.json`'s diff also includes benign, pre-existing `devOptional → dev` normalization on unrelated packages (reproducible on a clean `master` `npm install`; metadata-only, `npm ci`-safe), not caused by this change.

---

Deployed demo, note no report item UI shown now on the right hand side:

https://models-resources.concord.org/question-interactives/branch/remove-button-report-item/button/demo.html


emcelroy (approved):
Looks good 👍

